PR #5, opened 2019-01-29
Some fixes to convert keras application models

jiafatom:


jiafatom:
With this fix, we can convert all keras application models. For onnx runtime official 0.1.4, onnx runtime does not give correct prediction for NASNet. But it gives correct prediction when I built locally with config=RelWithDebInfo. Will file a bug for onnx runtime.

ketone/common/data_types.py
```diff
@@ -90,6 +90,14 @@ def _get_element_onnx_type(self):
         return onnx_proto.TensorProto.INT64
 
 
+class BoolTensorType(TensorType):
```
wenbingl: BooleanTensorType?

jiafatom: done

PR #6, opened 2019-01-30
Import onnxtk for optimize_onnx

jiafatom:


ketone/topology.py
```diff
@@ -251,7 +250,14 @@ def convert_topology(topology, model_name, doc_string, target_opset, channel_fir
 
 
     # enable the ONNX optimizations
-    nodes = optimize_onnx(container.nodes, nchw_inputs=nchw_inputs, inputs=container.inputs + extra_inputs, outputs=container.outputs)
+    try:
+        import onnxtk
+        nodes = onnxtk.optimizer.optimize_onnx(container.nodes, nchw_inputs=nchw_inputs, inputs=container.inputs + extra_inputs,
+                              outputs=container.outputs)
+    except ImportError:
+        ketone_logger().warning("onnxtk is not imported, so the convertor optimizer is not enabled, "
```
wenbingl: if nchw_input is not None, we should raise exception here, instead of warning.

PR #9, opened 2019-02-01
Instructions on running keras application models

jiafatom:


jiafatom:
Just copy the code from readme and it can run.

PR #16, opened 2019-02-08
Prepare to release the pypi package.

wenbingl:


requirements-dev.txt
```diff
@@ -7,3 +7,4 @@ onnx
 onnxtk
 pillow
 pytest
+onnxruntime
```
jiafatom: pip install will install onnxruntime 0.2.1 which will cause error right now

PR #20, opened 2019-02-10
Set softmax axis=-1 in Dense and add other activations

jiafatom:


jiafatom:
Synced offline, we also port the change https://github.com/onnx/onnxmltools/pull/217/files to add more activations to Dense layer.

PR #25, opened 2019-02-19
Create initializers using const op

jiafatom:


jiafatom:
Also set SoftMax axis=-1 (previously we only do it inside dense layer, now according to https://keras.io/activations/, we should make it consistent with keras io).

jiafatom:
This change cannot build for old tf2onnx 0.3.2, it works for the latest tf2onnx.

PR #26, opened 2019-02-19
Set default softmax axis=-1

jiafatom:
https://keras.io/activations/, keras.activations.softmax(x, axis=-1) 
-1 should be default.

### DIFF
--- a/keras2onnx/common/onnx_ops.py
+++ b/keras2onnx/common/onnx_ops.py
@@ -405,7 +405,7 @@ def apply_identity(scope, input_name, output_name, container, operator_name=None
     container.add_node('Identity', input_name, output_name, name=name)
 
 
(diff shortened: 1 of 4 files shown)


keras2onnx/common/onnx_ops.py
```diff
@@ -405,7 +405,7 @@ def apply_identity(scope, input_name, output_name, container, operator_name=None
     container.add_node('Identity', input_name, output_name, name=name)
 
 
-def apply_softmax(scope, input_name, output_name, container, operator_name=None, axis=1):
+def apply_softmax(scope, input_name, output_name, container, operator_name=None, axis=-1):
```
wenbingl: Are you sure -1 is also supported by ONNX?

jiafatom: will do this in conv.py

PR #27, opened 2019-02-21
bug fix for GRU converter and some code refinement.

wenbingl:


requirements.txt
```diff
@@ -1,7 +1,6 @@
 numpy
 protobuf
 tf2onnx
-tensorflow
```
jiafatom: cannot find module tensorflow, so we need this?

wenbingl: putting this line here will restrict the end user has to use tensorflow, but actually the use could use tensorflow-gpu as well.

jiafatom: we have build break, move this to requirements-dev?

PR #28, opened 2019-02-21
add back the dense layer translation.

wenbingl:


wenbingl:
> do we need a unit test for dense layer?

There are already two cases in test_layers.

PR #31, opened 2019-03-05
Fixes #30, opset 9 updates: PRelu, Flatten, MatMul, BatchNorm, Upsample

vinitra:
- ported over common op updates from onnxmltools for opset 9 (batchNorm, prelu, upsample)
- updated flatten op: bidirectional.py
- updated matmul op: dense.py
- ~~reorganized onnx_ops file in alphabetical order~~ (will be addressed in separate PR)
- ~~ported over crop common op [onnxmltools:249](https://github.com/onnx/onnxmltools/pull/249), deprecated from experimental op in opset 10~~  (will be addressed in separate PR)

PR #34, opened 2019-03-06
Update ONNX version to 1.4.1 in CI Build

vinitra:
* Update to ONNX v1.4.1 in CI Build scripts in accordance with opset 9 changes
* Fix apply_upsample inputs for opset 9 input check


vinitra:
The error is in regards to the BatchNormalization "spatial" operator, updated in #31 with opset 9 updates.

PR #37, opened 2019-03-07
Set default axis=-1 for softmax for keras

jiafatom:


keras2onnx/ke2onnx/dense.py
```diff
@@ -10,15 +10,7 @@
...
-
-_activation_map = {_get_activation('sigmoid'): apply_sigmoid,
-                   _get_activation('softmax'): apply_softmax,
-                   _get_activation('linear'): apply_identity,
-                   _get_activation('relu'): apply_relu,
-                   _get_activation('elu'): apply_elu,
-                   _get_activation('selu'): apply_selu,
-                   _get_activation('tanh'): apply_tanh,
-                   _get_activation('hard_sigmoid'): apply_hard_sigmoid}
+from .activation import _activation_map
```
wenbingl: If it was imported by module, it should be started with an underscore.

keras2onnx/ke2onnx/activation.py
```diff
@@ -4,7 +4,7 @@
 # license information.
 # --------------------------------------------------------------------------
 import keras
-from keras.activations import get as _get_activation
+from keras.activations import get as get_activation
```
wenbingl: it's not necessary to change the name after 'as', given this rename method is really a private stuff in this module.

keras2onnx/ke2onnx/dense.py
jiafatom: updated

keras2onnx/ke2onnx/activation.py
jiafatom: done

PR #38, opened 2019-03-08
Print detail debug info for onnxruntime test

jiafatom:


tests/test_layers.py
```diff
@@ -54,6 +55,26 @@ def run_onnx_runtime(self, case_name, onnx_model, data, expected, rtol=1.e-4, at
         if res and temp_model_file not in self.model_files:  # still keep the failed case files for the diagnosis.
             self.model_files.append(temp_model_file)
 
+        if not res:
```
wenbingl: when the model failed to convert or run with onnx-runtime, will this code block still be executed?

jiafatom: No, this code block only deals with the result mismatch. If model fails to convert or run with onnx-runtime, the exception at earlier would throw.

PR #40, opened 2019-03-13
Convert LPCNet model to onnx

jiafatom:


wenbingl:
No need to copy all python file from LPCNet, just give the script for the conversion and running with onnxruntime.

PR #50, opened 2019-04-02
Implement keras wrapper timedistributed 

jiafatom:


keras2onnx/_builtin.py
```diff
@@ -5,10 +5,19 @@
 ###############################################################################
 from .common.onnx_ops import apply_identity
 from .funcbook import set_converter
+from .ke2onnx.main import apply_reshape
```
wenbingl: why not import it as same as the first import line?

jiafatom: done

PR #51, opened 2019-04-02
Support the keras model created by tf.keras API.

wenbingl:


jiafatom:
Do we have some unit test for tf.keras API? Currently it seems the unit test only goes to keras but not tf.keras.

PR #62, opened 2019-04-11
onnxconverter_common code refactor (onnx_ops, data_types)

vinitra:
+ add onnxconverter_common package to the builds
+ replace common code with onnxconverter_common dependencies
+ note that onnxconverter_common onnx_ops files now covers all keras-onnx changes and all sklearn-onnx changes (https://github.com/onnx/onnxmltools/pull/279)


vinitra:
@wenbingl, @jiafatom: when you get a chance, can you review this PR?

keras2onnx/ke2onnx/upsample.py
```diff
@@ -47,10 +47,10 @@ def convert_keras_upsample(scope, operator, container, n_dims):
     # If channels_first is True, we don't need to permute the output of ONNX Upsample. Otherwise, similar to Crop's
     # conversion, a Transpose would be added.
     if channels_first:
-        apply_upsample(scope, input_tensor_name, operator.outputs[0].full_name, container, scales=scales)
+        apply_upsample(scope, [input_tensor_name], operator.outputs[0].full_name, container, scales=scales)
```
jiafatom: Why there is a change here from input_tensor_name to [input_tensor_name]? do we change interface somewhere?

keras2onnx/common/data_types.py
```diff
@@ -3,161 +3,5 @@
...
-    '''
-    Find the operator name for converting source_type into target_type
-    '''
-    if type(source_type) == type(target_type):
-        return 'identity'
-    elif type(target_type) == FloatTensorType:
-        return 'imageToFloatTensor'
-    else:
-        raise ValueError('Unsupported type conversion from %s to %s' % (source_type, target_type))
+from onnxconverter_common.data_types import *
```
jiafatom: Why not fully delete this file? Is it because we may put data types here that are only used by keras2onnx?

keras2onnx/common/onnx_ops.py
```diff
@@ -3,427 +3,5 @@
...
-        name = _create_name_or_use_existing_one(scope, 'Resize', operator_name)
-        attrs = {'name': name}
-        attrs['mode'] = mode.lower()
-
-        scales_tensor_name = scope.get_unique_variable_name(name + '_scales')
-        container.add_initializer(scales_tensor_name, onnx_proto.TensorProto.FLOAT, [len(scales)], scales)
-        inputs = [input_name, scales_tensor_name]
-        op_version = 10
-
-        container.add_node('Resize', inputs, output_name, op_version=op_version, **attrs)
```
jiafatom: Is onnxmltools in sync with the latest change of this file?

.azure-pipelines/linux-conda-CI.yml
```diff
@@ -42,6 +42,10 @@ jobs:
       pip install tensorflow
       pip install $(KERAS)
       pip install $(TF2ONNX)
+      git clone http://github.com/onnx/onnxmltools
```
jiafatom: Shall we change this to pip install after the package is published?

vinitra: Thanks for the review, @jiafatom! Yes, this PR is waiting for the package to be published.

keras2onnx/common/onnx_ops.py
vinitra: Yes, see: https://github.com/onnx/onnxmltools/pull/279

Every op has been confirmed, and it includes the upsample opset 10 changes.

keras2onnx/common/data_types.py
vinitra: Yes, and to maintain backwards compatibility.

keras2onnx/ke2onnx/upsample.py
vinitra: The apply_upsample function in onnxconverter_common/onnx_ops.py requires a input_name list, as "scales" is now an additional input in opset 9.

vinitra: See https://github.com/onnx/onnxmltools/pull/281

vinitra: Reverted Upsample API changes in latest update.

jiafatom: I am a little confused here. Previously this works without any changes on apply_upsample. Why we need make onnxmltools change to fix this? If we need this fix, why it works previously?

vinitra: There are two ways to do this: 
1. pass in input_name as a list containing a single string, where the upsample function uses that list as expected input.
2. pass in input_name as a string, where the upsample function accounts for this change in the function. The container.add_node function accounts for a string input and converts it to a list. 

Method 1 was implemented in the upsample function in apply_operation in ONNXMLTools (now onnx_ops in onnxconverter-common), and Method 2 was implemented in keras-onnx. To reconcile the difference, I originally aligned this version with method 1. After realizing most functions use method 2, I reverted the changes and aligned with the keras-onnx version. This required changing the original onnxmltools apply_upsample to require input_name as a string instead of a list. The code above is now outdated.

vinitra: Both methods work individually, but since we're moving all apply_ops to onnxconverter-common, using half of method 1 and half of method 2 will not work. The reason method 1 was even considered was due to the way the function was called in upsample.py from the coreml converter -- in opset 9, 'scale' became an additional input instead of an initializer, so in that case, a list of multiple string input names must be passed to container.add_node().

jiafatom: Just checked, for keras opset 9, input_name is not a list. Did I miss something?

jiafatom:
Please fix the build before check-in, other parts are good to go.

.azure-pipelines/linux-conda-CI.yml
vinitra: Resolved in latest update! Using pypi package: https://pypi.org/project/onnxconverter-common

PR #63, opened 2019-04-15
Add keras application unit tests

jiafatom:
It adds 6-7 minutes for azure pipeline unit test, which is acceptable.


jiafatom:
Use https://github.com/onnx/keras-onnx/pull/66 instead.

PR #64, opened 2019-04-15
Embedded the tf2onnx as a submodule.

wenbingl:


wenbingl:
> Does it mean that we always pull tf2onnx from latest GitHub not from pypi?

No, it sticks to a certain commit and not related to pypi stuff any more.

PR #65, opened 2019-04-15
Speed up azure pipeline by saving model only for failure case

jiafatom:
To remove unnecessary io, this PR will speed up azure pipeline.


jiafatom:
Since the model is small for unit test, don't see significant gain, close this.

PR #68, opened 2019-04-16
Disable vgg16 temporarily

jiafatom:
need debug why pytest fails while it passed locally.


jiafatom:
Use https://github.com/onnx/keras-onnx/pull/67

PR #71, opened 2019-04-22
Pick up the interpolation attributes of Keras's UpSampling2D layers.

kohei-us:
In ONNX, An Upsample operator may have the mode value of either 'nearest'
or 'linear'.  In Keras, UpSampling2D may have the interpolation value of
either 'nearest' or 'bilinear'.


keras2onnx/ke2onnx/upsample.py
```diff
@@ -29,6 +29,12 @@ def convert_keras_upsample(scope, operator, container, n_dims):
     else:
         raise ValueError('Unsupported dimension %s when converting Keras Upsampling layer' % n_dims)
 
+    mode = 'nearest'
+    if n_dims == 2:
```
jiafatom: I suggest that we do linear mode for various n_dims in this PR. If n_dims = 1 or 3, we also need handle linear mode.

keras2onnx/ke2onnx/upsample.py
```diff
@@ -29,6 +29,12 @@ def convert_keras_upsample(scope, operator, container, n_dims):
...
+        mode = op.interpolation
+        if mode == 'bilinear':
+            mode = 'linear'
```
jiafatom: mode='linear' if mode.endswith('linear') ? it can do 1-d, 2-d, 3-d.

keras2onnx/ke2onnx/upsample.py
kohei-us: I've made changes to remove the dimension check, to make it independent of the dimensions.

keras2onnx/ke2onnx/upsample.py
kohei-us: Agreed.  I've switched to using endswith() to generalize it a bit.

keras2onnx/ke2onnx/upsample.py
```diff
@@ -29,6 +29,12 @@ def convert_keras_upsample(scope, operator, container, n_dims):
     else:
         raise ValueError('Unsupported dimension %s when converting Keras Upsampling layer' % n_dims)
 
+    mode = 'nearest'
+    if hasattr(op, 'interpolation'):
+        mode = op.interpolation
+        if mode.endswith('linear'):
+            mode = 'linear'
```
jiafatom: mode = 'linear' if mode.endswith('linear') else op.interpolation

tests/test_layers.py
```diff
@@ -453,6 +453,8 @@ def test_upsample(self):
         ishape = (20, 20, 1)
         layer = keras.layers.UpSampling2D(size=(2, 3), data_format='channels_last')
         self._misc_conv_helper(layer, ishape)
+        layer = keras.layers.UpSampling2D(size=(2, 3), data_format='channels_last', interpolation='bilinear')
+        self._misc_conv_helper(layer, ishape)
```
jiafatom: please fix build failure, thanks

kohei-us:
Hmm... Obviously the version of Keras used in the linux-conda-ci agent does not support the interpolation argument in its UpSampling2D layer implementation...

jiafatom:
> Hmm... Obviously the version of Keras used in the linux-conda-ci agent does not support the interpolation argument in its UpSampling2D layer implementation...

Let's find since which keras version interpolation argument is supported. Then in unit test, we use
if StrictVersion(keras.__version__) >= StrictVersion()
to only enable this test for the keras version when interpolation argument is supported.

jiafatom:
Just checked https://github.com/keras-team/keras/releases, Add interpolation argument in UpSampling2D layer , in keras 2.2.3

kohei-us:
Thanks for your suggestion.  I've made changes to check the version of Keras to conditionally disable my test case.  Hopefully this will work.

tests/test_layers.py
```diff
@@ -454,6 +454,13 @@ def test_upsample(self):
         layer = keras.layers.UpSampling2D(size=(2, 3), data_format='channels_last')
         self._misc_conv_helper(layer, ishape)
 
+    @unittest.skipIf(StrictVersion(keras.__version__) < StrictVersion("2.2.3"),
+                     "The interpolation keyword not supported in Keras prior to 2.2.3.")
+    def test_upsample_bilinear(self):
```
jiafatom: Let's still put this in test_upsample, just use 
if StrictVersion(keras.__version__) < StrictVersion("2.2.3"):
       layer = keras.layers.UpSampling2D(size=(2, 3), data_format='channels_last', interpolation='bilinear')
       self._misc_conv_helper(layer, ishape)

keras2onnx/ke2onnx/upsample.py
jiafatom: resolve this?

jiafatom:
sign CLA please?

kohei-us:
> sign CLA please?

I need to check with someone in our organization before signing this.  Let me work on this, and I'll get back to you.

keras2onnx/ke2onnx/upsample.py
jiafatom: please rewrite this piece of code concisely like: mode = 'linear' if mode.endswith('linear') else op.interpolation

jiafatom:
Also please fix the inline comments about write statements concisely.

kohei-us:
> Also please fix the inline comments about write statements concisely.

Done.

keras2onnx/ke2onnx/upsample.py
kohei-us: Done.

jiafatom:
Since cla is stil pending, we apply the fix [here](https://github.com/onnx/keras-onnx/pull/101) instead. Thank you for your contribution.

kohei-us:
I know it's too late, but I've finally got approval internally to sign the CLA.

PR #85, opened 2019-05-08
Fix bidirectional output_seq=False case

jiafatom:


tests/test_layers.py
```diff
@@ -625,6 +625,22 @@ def test_LSTM_reshape(self):
         expected = model.predict(data)
         self.assertTrue(self.run_onnx_runtime('tf_lstm', onnx_model, data, expected))
 
+    def test_Bidirectional(self):
+        input_dim = 10
+        sequence_len = 5
+        from keras.layers import Bidirectional, Dense, LSTM, Activation
```
wenbingl: Don't use from ... import ... in the test case, which can't run in tf.keras

PR #96, opened 2019-05-28
[Opset 10] Updating ThresholdedRelu and onnxconverter-common package

vinitra:
+ Updates for ThresholdedRelu for graduation from experimental to full op
+ ~MaxPool and AveragePool have the ceil_mode attribute in Opset 10 -- default is 0, with floor in the calculation for spatial dimensions~ Covered in changes in PR #99
+ Adding unit test coverage for ThresholdedRelu
+ Updating onnxconverter_common version to 1.5.0

### DIFF
--- a/keras2onnx/common/onnx_ops.py
+++ b/keras2onnx/common/onnx_ops.py
@@ -5,3 +5,18 @@
 ###############################################################################
 
 from onnxconverter_common.onnx_ops import *
+
+def apply_thresholded_relu(scope, input_name, output_name, container, operator_name=None, alpha=None):
+    if alpha == None:
+        alpha = [1.0]
+
+    name = _create_name_or_use_existing_one(scope, 'ThresholdedRelu', operator_name)
+    attrs = {'name': name, 'alpha': alpha[0]}
+    if container.target_opset < 10:
+    # ThresholdedRelu graduated from an experimental op to a full op in opset 10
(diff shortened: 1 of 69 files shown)


keras2onnx/common/onnx_ops.py
```diff
@@ -5,3 +5,19 @@
 ###############################################################################
 
 from onnxconverter_common.onnx_ops import *
+from onnxconverter_common.onnx_ops import _create_name_or_use_existing_one
+
+def apply_thresholded_relu(scope, input_name, output_name, container, operator_name=None, alpha=None):
```
jiafatom: why not using apply_thresholded_relu in onnxconverter_common.onnx_ops ?

vinitra: The CI uses the published package version of onnxconverter_common, so it will have to wait for the new PyPi publish to use that version of the op. Does it make sense to wait for that instead?

jiafatom: It's better to let onnxconverter_common publish first, otherwise we will change this code again. Is there any other change pending? Otherwise we can publish it.

vinitra: Now that onnxconverter-common is published, this code in onnx_ops has been removed.

jiafatom:
Let's add tests for previous opsets to ensure test coverage. Thanks.

PR #103, opened 2019-06-05
Bug fix in forward and bi-directional lstm in handling bias

sonu1-p:
The converter was not setting the value of R bias correctly during
onnx conversion.

We simply set the R biases to zero since the current conversion code
is already setting W bias to the Keras' bias value. We expect the sum
of the biases Wb + Rb to be the same as Keras' bias since Keras uses
a single bias combining both.

Changes to be committed:

	modified:   keras2onnx/ke2onnx/bidirectional.py
	modified:   keras2onnx/ke2onnx/lstm.py
	modified:   tests/test_layers.py

### DIFF
--- a/keras2onnx/ke2onnx/bidirectional.py
+++ b/keras2onnx/ke2onnx/bidirectional.py
@@ -72,13 +72,13 @@ def convert_bidirectional(scope, operator, container):
         b = np.zeros(shape=(8, hidden_size))
         keras_b = forward_layer.get_weights()[2]  # This matrix is a concatenation of B[ifco] in Keras
         # Set up B_i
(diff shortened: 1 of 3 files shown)


jiafatom:
@sonu1-p  can you please sign CLA? thanks

sonu1-p:
Hi David, I have signed it, don't know why this page is not reflecting the same.

sonu1-p:
When I click on license/cla details, it says - You have signed the CLA for onnx/keras-onnx

sonu1-p:
Seems like it is reflected now. Thanks!

jiafatom:
LGTM, thank you very much for the change!

PR #104, opened 2019-06-06
Handle 2D input for BatchNormalization

jiafatom:


keras2onnx/ke2onnx/batch_norm.py
```diff
@@ -10,7 +10,7 @@
 
 def convert_keras_batch_normalization(scope, operator, container):
     op = operator.raw_operator
-    if op.axis != 3 and op.axis != -1:
+    if (op.axis != 3 and op.axis != -1) or len(op.input_shape) == 2:
         adjusted_input_name = operator.inputs[0].full_name
```
vinitra: Is the op.axis portion always false in the 2D case? The "or" might not let it reach the second part of the if staement.

jiafatom: yes, for 2d case, dimension 0 is batch, so the data is actually dimension 1. So there is no notion of channel at this point. So we should not apply transpose.

PR #105, opened 2019-06-06
Handle multiple dimension case for BatchNormalization

jiafatom:


jiafatom:
Will use [this](https://github.com/onnx/keras-onnx/pull/106)

PR #106, opened 2019-06-06
Handle multiple dimension case for BatchNormalization

jiafatom:


keras2onnx/ke2onnx/batch_norm.py
```diff
@@ -62,4 +69,4 @@ def convert_keras_batch_normalization(scope, operator, container):
                          momentum=momentum, spatial=spatial)
 
         # Permute [N,C,H,W] to [N,H,W,C]
```
vinitra: Clarify that this comment is for the 4D input shape case

keras2onnx/ke2onnx/batch_norm.py
```diff
@@ -10,11 +10,18 @@
 
 def convert_keras_batch_normalization(scope, operator, container):
     op = operator.raw_operator
-    if (op.axis != 3 and op.axis != -1) or len(op.input_shape) == 2:
+    skip_transpose = (op.axis != len(op.input_shape) - 1 and op.axis != -1) or len(op.input_shape) <= 2
+    if len(op.input_shape) > 2:
+        perm_1 = list(range(1, len(op.input_shape) - 1))
+        perm_1 = [0, len(op.input_shape) - 1] + perm_1
+        perm_2 = list(range(2, len(op.input_shape)))
+        perm_2 = [0] + perm_2 + [1]
```
vinitra: Also, might not ever need to calculate perm_1 and perm_2 if skip_transpose is False. Perhaps move these into the respective sections for converter efficiency?

jiafatom: thanks, will do

PR #116, opened 2019-06-11
Handle list input size in LSTM and bidirectional

jiafatom:


keras2onnx/ke2onnx/bidirectional.py
```diff
@@ -32,8 +32,11 @@ def convert_bidirectional(scope, operator, container):
     op = operator.raw_operator
     forward_layer = op.forward_layer
     backward_layer = op.backward_layer
-    input_size = op.input_shape[-1]
-    seq_length = op.input_shape[-2]
+    input_shape = op.get_input_shape_at(0)
```
vinitra: is there any unit test update for the bidirectional change?

jiafatom: I need find a test case for this which is not trivial. I add a TODO here and will test later.

jiafatom:
> Looks like there's a reshape dimensions error in the new unit test.

This error is because the test method does not handle order well, updated.

PR #117, opened 2019-06-11
Conversion of BatchNormalisation cannot deal with list `axis`

fmannhardt:
Fix issue with BatchNormalisation layer having a list `axis` not adding the required Transpose layer when converting a NHWC model. Maybe a regression of the fix here #92

PR #129, opened 2019-06-21
Fix for batchnorm attribute `axis` list case by adding transpose layer

vinitra:
Update batch norm list case as done in PR #117 
Add cosmetic changes to fit into code base

"Fix issue with BatchNormalisation layer having a list axis not adding the required Transpose layer when converting a NHWC model." - @fmannhardt

### DIFF
--- a/keras2onnx/ke2onnx/batch_norm.py
+++ b/keras2onnx/ke2onnx/batch_norm.py
@@ -18,12 +18,12 @@ def convert_keras_batch_normalization(scope, operator, container):
         raise AttributeError('There is no input_shape or _input_shape for the operator: ' + operator.full_name)
 
     if isinstance(op.axis, list):
-            if len(op.axis) == 1:
-                axis = op.axis[0]
-            else:
-                raise AttributeError('No support for more than one axis in: ' + operator.full_name)
+        if len(op.axis) == 1:
+            axis = op.axis[0]
         else:
-            axis = op.axis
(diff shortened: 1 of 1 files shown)


keras2onnx/ke2onnx/batch_norm.py
```diff
@@ -17,6 +17,14 @@ def convert_keras_batch_normalization(scope, operator, container):
     else:
         raise AttributeError('There is no input_shape or _input_shape for the operator: ' + operator.full_name)
 
+    if isinstance(op.axis, list):
+            if len(op.axis) == 1:
```
jiafatom: indent issue?

keras2onnx/ke2onnx/batch_norm.py
```diff
@@ -17,6 +17,14 @@ def convert_keras_batch_normalization(scope, operator, container):
...
+                axis = op.axis[0]
+            else:
+                raise AttributeError('No support for more than one axis in: ' + operator.full_name)
+        else:
```
jiafatom: indent issue?

keras2onnx/ke2onnx/batch_norm.py
vinitra: fixed!

keras2onnx/ke2onnx/batch_norm.py
vinitra: fixed!

PR #130, opened 2019-06-21
a couple of fixing for tf-dtype, logging and pep.

wenbingl:


wenbingl:
> Can we have unit test for these newly added conversions? We need ensure test coverage.
The layer unit test are already there, the only difference is keras own conversion instead of tf2onnx.

PR #133, opened 2019-06-26
Use tensorflow 1.13.1 for tf keras (not 1.14.0)

jiafatom:


jiafatom:
tf keras version does not work for tensorflow=1.14.0 because the node output_tensors can be a tensor not list. So let's support tensorflow==1.13.1

PR #140, opened 2019-07-02
Convert mask-rcnn to onnx

jiafatom:


jiafatom:
This PR has 1k+ commits, I am going to use [this](https://github.com/onnx/keras-onnx/pull/140) instead, essentially the same.

PR #145, opened 2019-07-03
Throw exception if a tf.keras model does not set TF_KERAS

jiafatom:


keras2onnx/main.py
```diff
@@ -75,6 +75,14 @@ def convert_keras(model, name=None, doc_string='', target_opset=None, channel_fi
     set_logger_level(logging.DEBUG if debug_mode else logging.INFO)
     tf2onnx.logging.set_level(logging.DEBUG if debug_mode else logging.INFO)
 
+    if isinstance(model, tf.keras.Model):
+        if not is_tf_keras:
+            k2o_logger().warning("WARNING: This is a tensorflow keras model, but keras standalone converter is used." +
+                                 " Please set environment variable TF_KERAS.")
```
vinitra: Specify TF_KERAS=1

keras2onnx/main.py
```diff
@@ -75,6 +75,14 @@ def convert_keras(model, name=None, doc_string='', target_opset=None, channel_fi
...
+    elif is_tf_keras:
+        k2o_logger().warning("WARNING: This is a keras standalone model, but tensorflow.keras converter is used." +
+                             " Please unset environment variable TF_KERAS.")
```
vinitra: Specify how to unset -- TF_KERAS=0?

jiafatom: set TF_KERAS=

keras2onnx/main.py
jiafatom: As long as the environment variable is set, then it will use tf.keras.

keras2onnx/main.py
```diff
@@ -75,6 +75,14 @@ def convert_keras(model, name=None, doc_string='', target_opset=None, channel_fi
     set_logger_level(logging.DEBUG if debug_mode else logging.INFO)
     tf2onnx.logging.set_level(logging.DEBUG if debug_mode else logging.INFO)
 
+    if isinstance(model, tf.keras.Model):
+        if not is_tf_keras:
+            k2o_logger().warning("WARNING: This is a tensorflow keras model, but keras standalone converter is used." +
+                                 " Please set environment variable TF_KERAS.")
+    elif is_tf_keras:
```
wenbingl: This never gonna happen.

keras2onnx/main.py
wenbingl: raise an exception since it can't proceed any further.

keras2onnx/main.py
jiafatom: It happens when this is a keras standalone model

Issue #146, opened 2019-07-04
Negative paddings


jojaii:
The converter is outputting negative paddings in the graph whenever "same" padding is encountered. The model I used is the minixception model (accessible [here](https://github.com/oarriaga/face_classification/blob/master/trained_models/emotion_models/fer2013_mini_XCEPTION.107-0.66.hdf5)).

This can be seen in the "conv2d_3" layer in the model. You may download the file [here](https://drive.google.com/file/d/1tJrFuYGVMwtDUYFkKQElRLDHOWN5c96C/view?usp=sharing).

Here is a screenshot of the layer when opened with [Netron](https://github.com/lutzroeder/netron):
![negative paddings 2](https://user-images.githubusercontent.com/42332973/60644510-ec508c80-9e68-11e9-85fe-bf3b1e79548e.png)

Here is the sample code:
```
import onnx
from keras.models import load_model
from keras2onnx import convert_keras

model = load_model('fer2013_mini_XCEPTION.107-0.66.hdf5')
onnx_model = convert_keras(model)
onnx.save(onnx_model, 'minixception_sample.onnx')
```

jiafatom:
@jojaii I have a fix for this, will check in soon.

jojaii:
Thanks @jiafatom. Will wait for the fix.

jiafatom:
> Thanks @jiafatom. Will wait for the fix.

@jojaii code checked in, thanks for reporting the issue.

PR #147, opened 2019-07-05
Fix padding=same for Conv

jiafatom:
To fix [issue](https://github.com/onnx/keras-onnx/issues/146)
We extract the core issue and add it as a unit test here:
```python
self._conv2_helper(8, 16, (1, 1), (2, 2), (60, 60), padding='same')
```
The original master code fails this unit test because it calculates a negative padding [-1, -1, 0, 0]. Now we fix it in this PR.

### DIFF
--- a/.azure-pipelines/linux-CI-keras-applications-nightly.yml
+++ b/.azure-pipelines/linux-CI-keras-applications-nightly.yml
@@ -14,9 +14,20 @@ jobs:
     vmImage: 'Ubuntu-16.04'
   strategy:
     matrix:
+      # No python 2.x since tf2onnx does not support it
+      Python35:
+        python.version: '3.5'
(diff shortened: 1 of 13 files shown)


wenbingl:
why not just fix the calculation? There is a history about not using auto_pad.

jiafatom:
> why not just fix the calculation? There is a history about not using auto_pad.

onnxruntime uses [this](https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/providers/cpu/nn/conv_base.h) to calculate output shape

Also refer to keras output shape computing [here](https://github.com/keras-team/keras/blob/b78e3ff9263a6c3b28c869a73f74a1fbd6fd491c/keras/utils/conv_utils.py#L85) function `def conv_output_length`. 

Will change our algorithm.

PR #154, opened 2019-07-15
update the yolo-v3 conversion code.

wenbingl:
simplify the logic and clean up some unnecessary code.


jiafatom:
Can we also fix mask-rcnn code about nms_node naming issue?

PR #158, opened 2019-07-16
Add Flatten to keras converter and fix _calc_explicit_padding in Conv

jiafatom:


wenbingl:
Add note in README.md to mention this GAN model has been verified now?

keras2onnx/ke2onnx/main.py
```diff
@@ -69,6 +69,28 @@ def convert_keras_concat(scope, operator, container):
                  operator_name=operator.full_name, axis=axis)
 
 
+def convert_keras_flatten(scope, operator, container):
```
themummy: Why not use the [ONNX Flatten Op](https://github.com/onnx/onnx/blob/master/docs/Operators.md#Flatten)?

PR #182, opened 2019-08-02
Add K.Any and K.All to builtin tf op conversion

jiafatom:
In tf2onnx conversion, there is a check
`utils.make_sure(all(i >= 0 for i in reduce_dim)`
in class AllAny, this is not needed.
Because (1) we have negative axis (2) ReduceSum and ReduceMin supports it.

### DIFF
--- a/keras2onnx/_builtin.py
+++ b/keras2onnx/_builtin.py
@@ -278,9 +278,36 @@ def on_TopKV2(ctx, node, name, args):
     node.type = "TopK"
 
 
+def on_AllAny(ctx, node, name, args):
+        # T output = All(T x, list(int) reduce_indices, @bool keepdims)
+        # T output = Any(T x, list(int) reduce_indices, @bool keepdims)
+        reduce_dim = node.inputs[1].get_tensor_value()
+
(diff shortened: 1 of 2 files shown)


keras2onnx/_builtin.py
```diff
@@ -278,9 +278,36 @@ def on_TopKV2(ctx, node, name, args):
...
+        ctx.make_node(op_type="Greater", inputs=[reduce_node.output[0], zero_node.output[0]],
+                      name=node.name, outputs=node.output, shapes=shapes, dtypes=dtypes)
+
+
 def tf2onnx_builtin_conversion(opset):
     return {
         'Round': (on_Round, []),
         'StridedSlice': (on_StridedSlice_9 if opset <= 9 else on_StridedSlice, []),
-        'TopKV2': (on_TopKV2, [])
+        'TopKV2': (on_TopKV2, []),
```
wenbingl: No ONNX version check?

jiafatom: No, ReduceSum and ReduceMin uses its onnx version 1 since.

PR #183, opened 2019-08-02
Support the model layer in time_distribution layer conversion.

wenbingl:


keras2onnx/common/interim.py
```diff
@@ -205,6 +205,20 @@ def add_initializer(self, name, onnx_type, shape, content):
         tensor = helper.make_tensor(name, onnx_type, shape, content)
         self.initializers.append(tensor)
 
+    def add_initializer_by_name(self, scope, original_name, onnx_type, shape, content):
```
jiafatom: In future, if we want to add initializer, should we use this function always?

jiafatom:
Please rebase and fix the build error

PR #189, opened 2019-08-07
Add apply functions into the OnnxOperatorBuilder.

wenbingl:


keras2onnx/proto/__init__.py
```diff
@@ -35,3 +36,11 @@ def _check_onnx_version():
     except ImportError:
         is_tf_keras = True
         from tensorflow.python import keras
+
+
+def is_keras_older_than(version_str):
```
jiafatom: Shall we cover the case "is_keras_older_than_or_equal_to"

wenbingl: can be achieved by not _later_than, don't dup it.

PR #191, opened 2019-08-08
Add onnx.checker.check_model in unit test

jiafatom:


jiafatom:
Don't rely on checker.

PR #193, opened 2019-08-08
Submodel name prefix matching when Dropout is skipped at K.set_learning_phase(0)

jiafatom:


keras2onnx/parser.py
```diff
@@ -257,17 +257,26 @@ def _convert_keras_sub_model(sub_model, graph, target_kenode, varset, top_kenode
...
+    name_set = []
+    sub_model_node_idx = 0
+    for nodes_ in sub_model._nodes_by_depth.values():
+        for n_ in nodes_:
+            for ts_ in n_.output_tensors:
+                name_set.append(ts_.name)
+
     if len(inbound_nodes) > 1 and inbound_nodes[0] is not target_kenode:
         # Assumption: the first node in the inbound node list is always the one used in the keras layers.
         base_node = inbound_nodes[0]
```
wenbingl: base_node can be removed in the new code and it should be gotten like sub_model._nodes_by_depth.get(sub_model_node_idx)

PR #199, opened 2019-08-14
Reuse run_onnx_runtime in unit tests

jiafatom:


tests/test_layers.py
```diff
@@ -16,6 +16,55 @@
 tmp_path = os.path.join(working_path, 'temp')
 
 
+def run_onnx_runtime(case_name, onnx_model, data, expected, model_files, rtol=1.e-3, atol=1.e-6):
```
wenbingl: It would be better to have a separated file for this function?

jiafatom: done.

PR #204, opened 2019-08-16
Add a command line for the converter.

wenbingl:


keras2onnx/cli.py
```diff
@@ -0,0 +1,43 @@
...
+import tensorflow as tf
+from .main import convert_tensorflow, convert_keras
+
+
+def main(input_file, output_file=None, inputs=None, outputs=None, opset=None, channel_first=None):
+    """
+    Provide a command line interface for the converter.
+    :param input_file: the original model file path
+    :param output_file: the converted ONNX model file path
+    :param inputs: The model graph input node list
```
jiafatom: inputs and outputs are only used for tensorflow graph, can we specify this clearly like:

:param inputs: The model graph input node list (tensorflow graph only)

PR #216, opened 2019-08-29
[test only] test Batchnormalization_3

jiafatom:


jiafatom:
Batchnormalization_3 passes with existing code.

PR #229, opened 2019-09-05
Fix submodel recursion, add cgan to unit test, other gans to application

jiafatom:


jiafatom:
depends on onnxconverter_common 1.5.4

PR #230, opened 2019-09-06
Add several GAN to applications

jiafatom:


.azure-pipelines/linux-conda-CI-tf-keras.yml
```diff
@@ -45,6 +45,7 @@ jobs:
       python -m pip install --upgrade pip
       conda config --set always_yes yes --set changeps1 no
       pip install $(ONNX_PATH)
+      pip install h5py==2.9.0
```
wenbingl: why does CI build needs this file too?

jiafatom: keras-applications depends on h5py. Today it released 2.10.0 and build wheel fails. so specify 2.9.0.

PR #237, opened 2019-09-13
Add yolov3 to nightly build

jiafatom:

PR #240, opened 2019-09-14
Changes in NMS layer converter to make shape of boxes_r consistent , …

amanarora28:
…yolo model is also added and AddV2 operations there are replaced with AddV1


wenbingl:
> Thanks for the change, this change can be seen as a work around for now.
> We are actively working on adding yolov3 to nightly build, see [here](https://github.com/onnx/keras-onnx/pull/237), and it needs other changes as well. It should work after the change finishes.
@amanarora28 , thanks for the contribution, and can you sign the CLA to merge your change. 

@jiafatom , the nightly build can integrate the changes in this PR.

wenbingl:
@amanarora28, any further thinking?

jiafatom:
Use [this](https://github.com/onnx/keras-onnx/pull/237)

PR #243, opened 2019-09-18
[test only] test keras 2.3.0

jiafatom:


jiafatom:
Use [this PR](https://github.com/onnx/keras-onnx/pull/244)

PR #263, opened 2019-10-02
Split mask-rcnn into two tests for nightly build

jiafatom:


jiafatom:
It still causes segmentation fault.

PR #272, opened 2019-10-17
Support causal padding for Conv1D

jiafatom:


keras2onnx/ke2onnx/conv.py
```diff
@@ -161,12 +162,26 @@ def convert_keras_conv_core(scope, operator, container, is_transpose, n_dims, in
...
+    elif op.padding == 'causal':
+        assert n_dims == 1
+        attrs['auto_pad'] = 'VALID'
+        left_pad = op.dilation_rate[0] * (op.kernel_size[0] - 1)
+        oopb = OnnxOperatorBuilder(container, scope)
+        padded_result = oopb.add_node('Pad',
+                                      [convolution_input_names[0],
+                                       ('_pad', oopb.int64, np.array([0, 0, left_pad, 0, 0, 0], dtype='int64'))],
+                                      operator.output_full_names[0] + '_padded_result',
+                                      op_domain='com.microsoft' if container.target_opset < 11 else '')
```
wenbingl: don't  use microsoft domain in the standard layer conversion.

jiafatom: done

keras2onnx/ke2onnx/conv.py
```diff
@@ -161,12 +162,22 @@ def convert_keras_conv_core(scope, operator, container, is_transpose, n_dims, in
                                                    op.dilation_rate,
                                                    list(range(
                                                        len(input_shape))) if channels_first else input_perm_axes)
+    elif op.padding == 'causal':
+        assert n_dims == 1
+        attrs['auto_pad'] = 'VALID'
+        left_pad = op.dilation_rate[0] * (op.kernel_size[0] - 1)
```
wenbingl: Are you sure the padding is only related to the kernel size, irrelevant to the input?

jiafatom: Yes, see the tensorflow backend [here](https://github.com/keras-team/keras/blob/master/keras/backend/tensorflow_backend.py#L3654). This pad is an extra one only used for causal mode, to do a preprocessing for causal. It is followed by Conv with another pad which is related to the kernel size

PR #275, opened 2019-10-21
test

jiafatom:


jiafatom:
There is indeed an onnxruntime bug introduced in ort-nightly 0.5.0.dev1015.

PR #283, opened 2019-10-30
Enable tf20 custom and lambda layer test cases.

wenbingl:


tests/test_layers.py
```diff
@@ -96,18 +96,15 @@ def test_keras_with_tf2onnx(self):
         expected = model.predict(data)
         self.assertTrue(run_onnx_runtime('ktf2onnx_test', onnx_model, data, expected, self.model_files))
 
-    @unittest.skipIf(is_tf2, 'lambda/custom layer conversion not ready')
     def test_keras_lambda(self):
         model = Sequential()
         model.add(Lambda(lambda x: x ** 2, input_shape=[3, 5]))
-        model.add(Lambda(lambda x: tf.round(x), input_shape=[3, 5]))
+        if get_opset_number_from_onnx() >= 11:
```
jiafatom: We support tf.round conversion for opset < 11, see [here](https://github.com/onnx/keras-onnx/blob/174bc9a30e7ea373b216e11ca3dd9dd024adc92b/keras2onnx/wrapper.py#L213). Let's enable this after finish apply_round.

PR #286, opened 2019-11-05
Convert tf.stridedslice, tf.round, tf.topkv2 by keras _builtin converter

jiafatom:


keras2onnx/_builtin.py
```diff
@@ -19,14 +20,19 @@ class TYPES:
     Any = 'Any'
     All = 'All'
     Cast = 'Cast'
+    StridedSlice = 'StridedSlice'
 
     # converter internal types:
     TD_Reshape = '_reshape_timedistributed'
 
 
+def _is_tensor_const(tensor):
```
wenbingl: Don't reply this test because in the future the converter probably retrieve the value from these node which could be const folding.

keras2onnx/_builtin.py
```diff
@@ -98,6 +104,147 @@ def convert_tf_case(scope, operator, container):
                               to=to)
 
 
+def process_begin_end(new_begin, new_end, stride):
```
wenbingl: No '_' for this private function?

PR #292, opened 2019-11-14
Remove unused initializers when making onnx model

jiafatom:


wenbingl:
It would be more general to implement it in the onnxconverter-common.

PR #298, opened 2019-11-20
Convert tf.stack, tf.tile, tf.min, etc via keras _builtin

jiafatom:


keras2onnx/_builtin.py
```diff
@@ -317,6 +394,18 @@ def convert_tf_squeeze(scope, operator, container):
                               axes=axis)
 
 
+@converter_func(TYPES.Tile)
+def convert_tf_tile(scope, operator, container):
+    oopb = OnnxOperatorBuilder(container, scope)
+    cast_1 = oopb.add_node('Cast',
```
vinitra: Why do we need a cast op here for tile?

tests/test_layers.py
```diff
@@ -193,6 +206,29 @@ def test_tf_squeeze(self):
             expected = model.predict(data)
             self.assertTrue(run_onnx_runtime('onnx_squeeze', onnx_model, data, expected, self.model_files))
 
+    def test_tf_stack(self):
+        def myFunc1(x):
```
vinitra: nit: consistent naming scheme with underscores, call this function test_func_1 or my_func_1

tests/test_layers.py
```diff
@@ -193,6 +206,29 @@ def test_tf_squeeze(self):
...
+            return tf.stack([x[0], x[1], x[2]], axis=1)
+
+        def myFunc2(x):
```
vinitra: likewise, my_func_2

keras2onnx/_builtin.py
jiafatom: onnx Tile input 1 needs to be int64.

PR #299, opened 2019-11-22
Convert tf.gather and tf.pad via keras _builtin

jiafatom:


keras2onnx/_builtin.py
```diff
@@ -174,6 +194,74 @@ def convert_tf_pack(scope, operator, container):
...
+
+    if operator.target_opset < 11:
+        attrs['pads'] = paddings
+        if mode in [None, "constant"] and len(node.inputs) == 3:
+            const_val = _cal_tensor_value(node.inputs[2]).tolist()
+            attrs['value'] = const_val
+        pad_node =  oopb.add_node("Pad",
+                                  cast_op,
+                                  name=operator.full_name + 'pad',
+                                  op_version=2, **attrs)
```
vinitra: Do we need to account for pad op_version=1 here?

jiafatom: I checked tf2onnx code, there is no difference between op_version=2 and op_version=1, we can do op_version=2 for now.

PR #301, opened 2019-12-06
Convert tf.range and dynamic shape case for tf.reshape

jiafatom:


keras2onnx/_builtin.py
```diff
@@ -140,6 +141,94 @@ def convert_tf_gather_v2(scope, operator, container):
...
+    trip_count_node = oopb.apply_cast(ceil_node,
+                                      to=onnx_proto.TensorProto.INT64,
+                                      name=operator.full_name + '_trip_cnt')
+    loop_inputs = [trip_count_node[0],
+                   # TENSOR_TYPE_TO_STORAGE_TENSOR_TYPE maps BOOL to INT32
+                   # so we need change np.array(True, dtype='bool') to int32 here
+                   ('_cond', onnx_proto.TensorProto.BOOL, np.array(1, dtype='int32')),
+                   start.name]
+    from onnx import helper
+    n1 = helper.make_node("Identity", ["cond"], ["cond_out"], name="n1")
```
wenbingl: should this be extracted as a function?

jiafatom: Since no other places use this structure, let's keep it for now until needed.

PR #308, opened 2019-12-12
Handle tensorflow.errors.OpError in _cal_tensor_value and fix get_input/output_shape

jiafatom:


keras2onnx/common/intop.py
```diff
@@ -89,13 +89,23 @@ def get_attr(self, key):
         return self.attrs.get(key, None)
 
     def get_input_shape(self):
-        input_shape = self.inputs[0].type.shape
+        input_shape = [ None if i == 'N' else i for i in self.inputs[0].type.shape ]
```
wenbingl: if is is instance of string?

keras2onnx/common/intop.py
```diff
@@ -89,13 +89,23 @@ def get_attr(self, key):
...
+            if num_None_1 > num_None_2:
+                input_shape = self.raw_operator.output_shape
         return input_shape
 
     def get_output_shape(self):
-        output_shape = self.outputs[0].type.shape
+        output_shape = [ None if i == 'N' else i for i in self.outputs[0].type.shape ]
         if output_shape is None:
             output_shape = self.raw_operator.output_shape
+        elif hasattr(self.raw_operator, 'output_shape'):
```
wenbingl: Need update parser code to enable type.shape is as same as the layer.input/output shape

keras2onnx/common/intop.py
jiafatom: done

PR #311, opened 2019-12-13
Fixing the issue of Masking layer with a custom layer.

wenbingl:


jiafatom:
Please fix build before check in.

PR #327, opened 2019-12-26
Convert tf.unstack via keras _builtin

jiafatom:


keras2onnx/_builtin.py
```diff
@@ -978,6 +979,29 @@ def convert_tf_strided_slice(scope, operator, container):
                                   operator.inputs[0].full_name + '_identity')
 
 
+@converter_func(TYPES.Unpack)
+def convert_tf_unpack(scope, operator, container):
+    oopb = OnnxOperatorBuilder(container, scope)
+    node = operator.raw_operator
+    axis_val = node.get_attr('axis')
+    input_shape = _cal_tensor_shape(node.inputs[0])
+    if axis_val < 0 and operator.target_opset < 11:
```
vinitra: Can you explain the case where axis_val is negative? Is it the default value? Might be useful to have a comment here.

keras2onnx/_builtin.py
```diff
@@ -978,6 +979,29 @@ def convert_tf_strided_slice(scope, operator, container):
...
+                                   operator.inputs[0].full_name,
+                                   operator.full_name + '_split',
+                                   outputs_num=input_shape[axis_val],
+                                   axis=axis_val)
+
+    for i in range(len(split_node)):
+        oopb.apply_op_with_output("apply_squeeze",
+                                  split_node[i],
+                                  operator.outputs[i].full_name,
+                                  name=operator.full_name + '_squeeze' + str(i),
```
vinitra: Should there be an underscore (_) between squeeze and str(i)? If this is consistent style with other functions, feel free to disregard.

tests/test_layers.py
```diff
@@ -463,6 +463,15 @@ def test_tf_tile(self):
...
         self.assertTrue(run_onnx_runtime('onnx_tile', onnx_model, data, expected, self.model_files))
 
+    def test_tf_unstack(self):
+        for axis in [1, -1]:
+            model = Sequential()
+            model.add(Lambda(lambda x: tf.unstack(x, axis=axis)[0], input_shape=[2, 3, 4]))
+            onnx_model = keras2onnx.convert_keras(model, 'test_tf_unstack')
+            data = np.random.rand(3, 2, 3, 4).astype(np.float32)
+            expected = model.predict(data)
+            self.assertTrue(run_onnx_runtime('onnx_unstack', onnx_model, data, expected, self.model_files))
```
vinitra: Should this be onnx_unpack for clarity?

keras2onnx/_builtin.py
jiafatom: axis_val can be negative when the user wants to specify the axis counting back from the last dimension. Negative axis is commonly used.

keras2onnx/_builtin.py
jiafatom: Good catch, will add underscore there.

tests/test_layers.py
jiafatom: tensorflow unpack is an old op which is merged into tf unstack. In API, we are using tf.unstack, this is why I use onnx_unstack here. The underlying op is tf unpack. I can use unpack instead. thanks.

PR #328, opened 2019-12-27
Convert tf.transpose via keras _builtin

jiafatom:


tests/test_layers.py
```diff
@@ -463,6 +463,32 @@ def test_tf_tile(self):
...
+        data = np.random.rand(4, 2, 3, 4).astype(np.float32)
+        expected = model.predict(data)
+        self.assertTrue(run_onnx_runtime('onnx_transpose_2', onnx_model, data, expected, self.model_files))
+
+        def my_func_1(x):
+            a = tf.constant([[1, 2, 3], [4, 5, 6]], tf.float32)
+            return x + tf.transpose(a)
+
+        model = Sequential()
+        model.add(Lambda(lambda x: my_func_1(x), input_shape=[3, 2]))
```
vinitra: why not just pass in my_func_1 instead of the lambda function?

jiafatom: Will do next time

PR #334, opened 2020-01-02
[test] gemm test

jiafatom:


jiafatom:
We confirm the issue is because pipeline failure of ort-nightly. Fixed on onnxruntime side.

PR #335, opened 2020-01-07
Convert tf.VariableV2 via keras _builtin

jiafatom:


keras2onnx/_parse_tf.py
```diff
@@ -33,7 +33,7 @@ def infer_variable_type(tensor, opset, inbound_node_shape=None):
         tensor_shape = list(inbound_node_shape)
 
     # Determine the tensor's element type
-    tensor_type = tensor.dtype
+    tensor_type = tensor.dtype.base_dtype
```
wenbingl: compatible with between 1.x and 2.x?

PR #343, opened 2020-01-13
Convert tf.where via keras _builtin

jiafatom:


keras2onnx/proto/__init__.py
```diff
@@ -47,11 +47,12 @@ def is_keras_later_than(version_str):
     return StrictVersion(keras.__version__.split('-')[0]) > StrictVersion(version_str)
 
 
-if is_tf_keras:
+try:
```
wenbingl: No need to try, tensorflow existence will be check on converter startup.

PR #350, opened 2020-01-16
support transforms bert model conversion

wenbingl:


keras2onnx/_builtin.py
```diff
@@ -83,11 +97,35 @@ def _is_nhwc(node):
     return node.get_attr('data_format') == b'NHWC'
 
 
-def _cal_tensor_value(tensor):  # type: (tensorflow.Tensor)->np.ndarray
-    node = tensor.op
-    if node.type in ['Placeholder']:
+_MAX_FOLDING_NODE_NUMBER = 9
```
jiafatom: Is there a particular reason to choose 9 here?

wenbingl: No, empirical number.

PR #360, opened 2020-01-22
[Investigation] BatchNormalization output error, check ort-nightly 0119

jiafatom:


jiafatom:
With a fix on onnxruntime and ort-nightly update, nightly build runs fine.

PR #361, opened 2020-01-24
Convert tf.greater_equal, less_equal, scatter_nd via keras _builtin

jiafatom:
May add unit test for scatter_nd later.

### DIFF
--- a/keras2onnx/_builtin.py
+++ b/keras2onnx/_builtin.py
@@ -36,6 +36,8 @@ class TYPES:
     FusedBatchNormV3 = 'FusedBatchNormV3'
     GatherNd = 'GatherNd'
     GatherV2 = 'GatherV2'
+    GreaterEqual = 'GreaterEqual'
+    LessEqual = 'LessEqual'
     MatMul = 'MatMul'
     Max = 'Max'
     Maximum = 'Maximum'
@@ -56,6 +58,7 @@ class TYPES:
     ResizeNearestNeighbor = 'ResizeNearestNeighbor'
     Round = 'Round'
     Rsqrt = 'Rsqrt'
+    ScatterNd = 'ScatterNd'
     Select = 'Select'
     Shape = 'Shape'
     Size = 'Size'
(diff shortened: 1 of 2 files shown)


keras2onnx/_builtin.py
```diff
@@ -524,6 +527,40 @@ def convert_tf_gather_nd(scope, operator, container):
...
 
+def _convert_tf_compare_equal(scope, operator, container, tf_op_string, onnx_op_string):
+    if operator.target_opset < 7:
+        raise ValueError(tf_op_string + " op is not supported for opset < 7")
+    oopb = OnnxOperatorBuilder(container, scope)
+    if operator.target_opset >= 9:
+        compare_node = oopb.add_node(onnx_op_string,
+                                     operator.input_full_names,
+                                     operator.full_name + '_' + onnx_op_string.lower())
+        oopb.add_node_with_output('Not',
```
vinitra: I assume this "Not" is to address the not >= or the <= case, but why is it added to the graph in every case? Shouldn't it be dependent on the result of the comparison?

keras2onnx/_builtin.py
```diff
@@ -924,6 +961,43 @@ def convert_tf_reshape(scope, operator, container):
                               desired_shape=shape_value)
 
 
+@converter_func(TYPES.ScatterNd)
+def convert_tf_scatter_nd(scope, operator, container):
+    if operator.target_opset < 11:
+        raise ValueError("ScatterNd op is not supported for opset = " + str(operator.target_opset))
```
vinitra: To stay consistent w/ other functions (i.e. line 532), should be "... not supported for opset < 11"

keras2onnx/_builtin.py
```diff
@@ -924,6 +961,43 @@ def convert_tf_reshape(scope, operator, container):
...
+
+        node_input_0_dtype = _to_onnx_type(node.inputs[0].dtype)
+        if node_input_0_dtype != oopb.int64:
+            node_input_0_cast = oopb.apply_cast(operator.inputs[0].full_name,
+                                                to=oopb.int64,
+                                                name=operator.full_name + '_input_0')
+        else:
+            node_input_0_cast = [operator.inputs[0].full_name]
+
+        oopb.add_node_with_output('ScatterND',
```
vinitra: 'ScatterNd' vs. 'ScatterND'

tests/test_layers.py
```diff
@@ -332,18 +332,19 @@ def my_func_3(x):
...
+            input1 = Input(shape=input1_shape[0], dtype='int32')
+            input2 = Input(shape=input1_shape[1], dtype='int32')
+            comp = Lambda(lambda x: tf_op_(x[0], x[1]))([input1, input2])
+            model = keras.models.Model(inputs=[input1, input2], outputs=comp)
+
+            onnx_model = keras2onnx.convert_keras(model, 'tf_compare_equal')
+            data1 = np.array([[1, 2, 3], [1, 2, 3]]).astype(np.int32)
+            data2 = np.array([[1, 2, 3], [2, 1, 4]]).astype(np.int32)
+            expected = model.predict([data1, data2])
+            self.assertTrue(run_onnx_runtime('tf_compare_equal', onnx_model, [data1, data2], expected, self.model_files))
```
vinitra: test for scatter_nd?

vinitra: Nevermind, I see your comment above re: adding it later.

keras2onnx/_builtin.py
jiafatom: GreaterEqual = Less + Not

keras2onnx/_builtin.py
vinitra: Addressed offline. GreaterEqual = Less + Not and vice versa.

keras2onnx/_builtin.py
jiafatom: op name: tf ScatterNd, but onnx ScatterND

PR #370, opened 2020-02-06
Add tf.cast to test_tf_slice in unit test

jiafatom:


jiafatom:
Just realize PR 369 did this..so this change is not necessary. Just remove one line.

PR #386, opened 2020-02-17
Masking RNN with zeros input

cjermain:
This PR reproduces a discrepancy between the ONNX and Keras models for handling inputs that are fully masked. This occurs for RNN models that contain both a masking layer, and use a non-zero bias. When one of the input entries is fully zero (i.e. will be fully masked by the Masking layer), the RNN output is non-zero. Turning off the bias or using the default bias reproduces consistent behavior with Keras.

```
(Pdb) up
> /home/user/keras-onnx/tests/test_layers.py(1817)test_masking_bias()
-> self.assertTrue(run_onnx_runtime(onnx_model.graph.name, onnx_model, x, expected, self.model_files))
(Pdb) x
array([[[365.8103 , 107.39152, 701.68   , 615.55023, 468.88876],
        [831.7788 , 680.243  , 890.5216 , 550.38116, 970.3993 ],
        [134.01295, 771.0398 , 594.99915, 315.76196, 753.1429 ]],

       [[  0.     ,   0.     ,   0.     ,   0.     ,   0.     ],
        [  0.     ,   0.     ,   0.     ,   0.     ,   0.     ],
        [  0.     ,   0.     ,   0.     ,   0.     ,   0.     ]]],
      dtype=float32)
(Pdb) down
> /home/user/keras-onnx/tests/test_utils.py(153)run_onnx_runtime()
-> return res
(Pdb) expected
[array([[ 0.7615942,  0.       , -0.7615942,  0.       ,  0.       ,
         0.       ,  0.7615942,  0.       ],
       [ 0.       ,  0.       ,  0.       ,  0.       ,  0.       ,
         0.       ,  0.       ,  0.       ]], dtype=float32)]
(Pdb) actual
[array([[ 0.76159436,  0.        , -0.76159436,  0.        ,  0.        ,
         0.        ,  0.76159436,  0.        ],
       [ 0.2423832 ,  0.26090473,  0.08070117,  0.30603507,  0.13741694,
         0.24133024,  0.513518  ,  0.37147206]], dtype=float32)]
```

This is a problem for architectures using multiple RNN layers with masking. Any suggestions or thoughts are appreciated!

### DIFF
--- a/keras2onnx/ke2onnx/simplernn.py
+++ b/keras2onnx/ke2onnx/simplernn.py
@@ -85,7 +85,7 @@ def convert_keras_simple_rnn(scope, operator, container):
     oopb = OnnxOperatorBuilder(container, scope)
 
     if uses_masking_layer:
-        mask_cast = oopb.add_node('Cast', operator.input_masks[0].full_name, operator.full_name + '_mask_cast', to=oopb.int32)
+        mask_cast = oopb.apply_cast(operator.input_masks[0].full_name, to=oopb.int32, name=operator.full_name + '_mask_cast')
         oopb.add_node_with_output('ReduceSum', mask_cast, sequence_lengths, keepdims=False, axes=[-1], name=operator.full_name + '_mask_sum')
 
     oopb.apply_op_with_output('apply_rnn',
--- a/keras2onnx/ke2onnx/simplernn.py
+++ b/keras2onnx/ke2onnx/simplernn.py
@@ -48,7 +48,13 @@ def convert_keras_simple_rnn(scope, operator, container):
         rnn_input_names.append('')
 
     # sequence_lens is not able to be converted from input_length
-    rnn_input_names.append('')
+    uses_masking_layer = len(operator.input_masks) == 1
+    if uses_masking_layer:
+        # Mask using sequence_lens input
+        sequence_lengths = scope.get_unique_variable_name(operator.full_name + '_seq_lens')
+        rnn_input_names.append(sequence_lengths)
+    else:
+        rnn_input_names.append('')
     # inital_h: none
     if len(operator.inputs) == 1:
         rnn_input_names.append('')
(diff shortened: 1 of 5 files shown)


cjermain:
The following is the ONNX graph in Netron. It seems like any masked rows are set to zeros in the input to the LSTM (at the Mult operator). I would have expected a operator after the LSTM to set any masked values to their corresponding value (i.e. zero in this case).
temp onnx

cjermain:
I've added a section to generate the `sequence_lens` input for the [LSTM operator](https://github.com/onnx/onnx/blob/master/docs/Operators.md#LSTM), based on the output mask from the `Masking` layer. This replicates the functionality in Keras, and fixes the blocking test that I had added earlier.

temp_sequential onnx

cjermain:
This same strategy should be able to be applied to `bidirectional.py`. I will try that next.

wenbingl:
ready for reviewing or still WIP?

cjermain:
@wenbingl it would be great to get a review. I am thinking to hold off on the bidirectional implementation as a separate PR.

keras2onnx/ke2onnx/simplernn.py
```diff
@@ -77,6 +83,11 @@ def convert_keras_simple_rnn(scope, operator, container):
     rnn_output_names.append(rnn_y_name)
     rnn_output_names.append(rnn_h_name)
     oopb = OnnxOperatorBuilder(container, scope)
+
+    if uses_masking_layer:
+        mask_cast = oopb.add_node('Cast', operator.input_masks[0].full_name, operator.full_name + '_mask_cast', to=oopb.int32)
```
wenbingl: Can you replace add_node with oopb.apply_cast(...)? I know you probably was mislead by some bad examples that we haven't fixed yet. but with apply_cast function set, the converter will be transparent with the different opset.

keras2onnx/ke2onnx/simplernn.py
```diff
@@ -48,7 +48,13 @@ def convert_keras_simple_rnn(scope, operator, container):
         rnn_input_names.append('')
 
     # sequence_lens is not able to be converted from input_length
-    rnn_input_names.append('')
+    uses_masking_layer = len(operator.input_masks) == 1
+    if uses_masking_layer:
+        # Mask using sequence_lens input
+        sequence_lengths = scope.get_unique_variable_name(operator.full_name + '_seq_lens')
```
wenbingl: where is this tensor fed from the input masking data? do I miss something?

keras2onnx/ke2onnx/simplernn.py
cjermain: I've made the replacement.

keras2onnx/ke2onnx/simplernn.py
cjermain: The `sequence_lengths` variable defined here, which is passed to the LSTM layer, is used below to connect the `Masking` layer output. 
https://github.com/onnx/keras-onnx/pull/386/files#diff-e1267cefb3f38c599bf90778bb0b4dd1R89

PR #389, opened 2020-02-18
Custom Masking value

cjermain:
This PR uses the actual `mask_value`, instead of assuming that it is `0.0`. This allows custom mask values to be used, as long as they can be represented as a float32 or int32 (e.g. `np.nan` works as expected).

I added `mask_value` to the  `Operator` class to be able to pass that information down. I'm not convinced this is the best form, but it seemed to be the easiest way in the short term.

### DIFF
--- a/keras2onnx/ke2onnx/main.py
+++ b/keras2onnx/ke2onnx/main.py
@@ -68,9 +68,8 @@ def convert_keras_flatten(scope, operator, container):
 
 
 def _apply_not_equal(oopb, target_opset, operator):
-    mask_value = np.array([operator.mask_value], dtype='float32')
     if target_opset >= 11:
-        equal_out = oopb.add_node('Equal', [operator.inputs[0].full_name, mask_value],
+        equal_out = oopb.add_node('Equal', [operator.inputs[0].full_name, np.array([operator.mask_value], dtype='float32')],
                                   operator.full_name + 'mask')
         not_o = oopb.add_node('Not', equal_out,
                               name=operator.full_name + '_not')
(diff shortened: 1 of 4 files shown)


keras2onnx/ke2onnx/main.py
```diff
@@ -78,7 +79,7 @@ def _apply_not_equal(oopb, target_opset, operator):
                              "the masking layer result may be incorrect if the model input is in range (0, 1.0).")
         equal_input_0 = oopb.add_node('Cast', [operator.inputs[0].full_name],
                                       operator.full_name + '_input_cast', to=6)
-        equal_out = oopb.add_node('Equal', [equal_input_0, np.array([0], dtype='int32')],
```
cjermain: It looks like this should actually be `int32`, so the dtype is wrong.

keras2onnx/ke2onnx/main.py
```diff
@@ -69,7 +69,7 @@ def convert_keras_flatten(scope, operator, container):
 
 def _apply_not_equal(oopb, target_opset, operator):
     if target_opset >= 11:
-        equal_out = oopb.add_node('Equal', [operator.inputs[0].full_name, np.array([0], dtype='float32')],
+        equal_out = oopb.add_node('Equal', [operator.inputs[0].full_name, np.array([operator.mask_value], dtype='float32')],
```
wenbingl: In some case, mask_value will be None?

cjermain: Fair point. I raised an exception in this case, because it will only occur if the parsing does not work. I think it is preferable to have a clear error, than to silently use `0.0` as the default.

PR #395, opened 2020-02-26
Add vgg16 and nasnet to tfv2 application

jiafatom:


applications/nightly_build/test_keras_applications_v2.py
```diff
@@ -39,15 +45,21 @@ def test_MobileNetV2(self):
         res = run_image(model, self.model_files, img_path, tf_v2=True)
         self.assertTrue(*res)
 
+    def test_NASNetMobile(self):
+        NASNetMobile = keras.applications.nasnet.NASNetMobile
+        model = NASNetMobile(weights='imagenet')
```
wenbingl: can we just test on the model without pre-trained weights to save some time from downloading data.

PR #405, opened 2020-03-08
Refactor RNN parameter extracting

cjermain:
This PR is a step towards supporting SimpleRNN and GRU in the Bidirectional layer, by moving toward a consistent API across the RNN layers. It refactors the RNN code to increase sharing, and re-uses that in the Bidirectional LSTM case. No functional changes are made -- this PR is purely a refactor.

@wenbingl, can you review these changes?


wenbingl:
Thanks a lot for making RNN converter better!

PR #406, opened 2020-03-09
Convert tf.add_n using keras _builtin

jiafatom:


tests/test_layers.py
```diff
@@ -97,6 +97,21 @@ def test_keras_lambda(self):
         expected = model.predict(data)
         self.assertTrue(run_onnx_runtime('onnx_lambda', onnx_model, data, expected, self.model_files))
 
+    def test_tf_addn(self):
+        input1 = Input(shape=(5, 3, 4), dtype=tf.float32)
+        input2 = Input(shape=(5, 3, 4), dtype=tf.float32)
+        added = Lambda(tf.math.add_n)([input1, input2])
```
vinitra: maybe call this "sum" instead of "added" for clarity

PR #417, opened 2020-03-29
Support initial states for Bidirectional RNN

cjermain:
This PR provides support for initial states in Bidirectional SimpleRNN, GRU, and LSTM layers. It also fixes a bug where leveraging two outputs of a Keras node can cause parsing failures (see the changes in `parser.py`). Support was extended for LSTM and Bidirectional layers that have dynamic sequence lengths (`seq_length = None`), so that it now supports down to opset 5. This was previously possible in terms of the ONNX spec, but would raise an error.

### DIFF
--- a/keras2onnx/parser.py
+++ b/keras2onnx/parser.py
@@ -20,6 +20,9 @@ from ._parser_1x import (extract_inbound_nodes,
                          list_input_tensors, list_input_mask, list_output_mask,
                          list_output_tensors, list_input_shapes, list_output_shapes, on_parsing_keras_layer)
 
+ALLOWED_SHARED_KERAS_TYPES = {
+    keras.layers.embeddings.Embedding,
+}
 
 def _find_node(nodes, name):
     try:
@@ -570,6 +573,7 @@ def _parse_graph_core(graph, keras_node_dict, topology, top_scope, output_names)
     for n_ in model_outputs:
         q_overall.put_nowait(n_)
 
+    visited_layers = set()
     visited = set()  # since the output could be shared among the successor nodes.
     inference_nodeset = _build_inference_nodeset(graph, model_outputs)
     keras_nodeset = _build_keras_nodeset(inference_nodeset, keras_node_dict)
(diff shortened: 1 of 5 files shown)


cjermain:
Looks like there are issues in Python 3.7 that need to be addressed.

cjermain:
@wenbingl, can you review this PR?

keras2onnx/parser.py
```diff
@@ -581,6 +585,14 @@ def _parse_graph_core(graph, keras_node_dict, topology, top_scope, output_names)
         nodes = []
         layer_key_, model_ = _parse_nodes(graph, inference_nodeset, input_nodes, keras_node_dict, keras_nodeset,
                                           node, nodes, varset, visited, q_overall)
+
+        # Only parse Keras layers once (allow certain shared classes)
+        if layer_key_ in visited_layers:
```
wenbingl: In my understanding, are these lines only for an optimization, not for a bug fixing, and it looks any layers without weight could be ALLOWED_SHARED_KERAS_TYPES, right?

cjermain: The issue arose when I tied 2 LSTM layers together in `test_layers.py:test_rnn_state_passing`. Since there are 2 state outputs (hidden and cell) from the 1st LSTM to the 2nd LSTM, it winds up creating a duplicate object for the 1st LSTM and failing. This is because it does not correctly recognize that the 1st LSTM layer has already been considered, because the hidden output node is left over (not present in `visited`) when the cell output is traversed. I found it easier to fix the issue by ignoring any layer that had already been processed, but provided a mechanism to handle cases when this was desirable. I am not that familiar with this part of the code, so there may be a more optimal solution.

jiafatom: This change will break mask_rcnn in existing nightly build, see [here](https://dev.azure.com/onnxmltools/ketone/_build/results?buildId=11028&view=logs&j=b5dc3593-0456-5043-06cc-36317977280a)

jiafatom: Fixed [here](https://github.com/onnx/keras-onnx/pull/427)

PR #421, opened 2020-04-01
Support masking for tf2 and tf.keras

jiafatom:


keras2onnx/_parse_tf.py
```diff
@@ -269,14 +269,28 @@ def build_layer_output_from_model(model, output_dict, input_names, output_names)
         return graph
 
 
+def _get_layer_endpoints(layer_endpoints, layer_info_end_points):
```
wenbingl, line 272: Can you leave some comments about this function?

PR #422, opened 2020-04-02
Support RNN for tf2 and tf.keras

jiafatom:


keras2onnx/_parse_tf.py
```diff
@@ -287,7 +287,8 @@ def on_parsing_keras_layer_v2(graph, layer_info, varset, prefix=None):
     operator = varset.declare_local_operator(type(layer), raw_model=layer, op_name=layer.name)
     operator.nodelist = node_list
 
-    if hasattr(layer, 'input_mask') and layer.input_mask is not None:
+    input_candidates = layer.input if isinstance(layer.input, list) else [layer.input]
+    if len(input_candidates) != len(layer_info.inputs):
```
wenbingl, line 291: Not sure if it is reasonable here?

jiafatom: looks reasonable for RNN case

PR #425, opened 2020-04-06
unittest -> pytest

cjermain:
This PR migrates the testing from `unittest` to `pytest`. This allows for the use of parametrized tests and fixtures, leading to more robust test coverage and easier test-code maintenance. This was motivated by my recent PRs on Bidirectional and Masking features, where I saw that many of the tests rely on nested for-loops, which are hard to filter (using `-k`) and maintain as they expand.

Goals:
- [ ] Replace nested for-loops with parametrized tests
- [ ] Replace test running functions with fixtures
- [ ] Improve access to `actual` predictions and better integrate into pytest/pdb
- [ ] Expand test coverage with additional parametrized tests

This PR is still a work in progress, and will be updated. I'm interested in any suggestions or comments.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,32 @@
+###############################################################################
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+###############################################################################
+import os
+import pytest
+
+from keras2onnx.proto import keras
+from test_utils import run_onnx_runtime
+
+K = keras.backend
+
+
+@pytest.fixture(scope='function')
+def runner():
+    model_files = []
+
+    def runner_func(*args, **kwargs):
+        return run_onnx_runtime(*args, model_files, **kwargs)
+
+    # Ensure Keras layer naming is reset for each function
+    K.reset_uids()
+    # Reset the TensorFlow session to avoid resource leaking between tests
+    K.clear_session()
+
+    # Provide wrapped run_onnx_runtime function
+    yield runner_func
+
+    # Remove model files
+    for fl in model_files:
+        os.remove(fl)
--- a/tests/test_cgan.py
+++ b/tests/test_cgan.py
@@ -120,7 +120,7 @@ class CGAN():
 
 
 @pytest.mark.skipif(keras2onnx.proto.tfcompat.is_tf2 and is_tf_keras, reason="Tensorflow 1.x only tests.")
-@pytest.mark.skipif(is_tf_keras and StrictVersion(tf.__version__) < StrictVersion("1.14.0"),
+@pytest.mark.skipif(is_tf_keras and StrictVersion(tf.__version__.split('-')[0]) < StrictVersion("1.14.0"),
                  reason="Not supported before tensorflow 1.14.0 for tf_keras")
 def test_CGAN(runner):
     keras_model = CGAN().combined
--- a/tests/test_cgan.py
+++ b/tests/test_cgan.py
@@ -3,8 +3,7 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 ###############################################################################
-import os
-import unittest
+import pytest
 import tensorflow as tf
 import keras2onnx
 import numpy as np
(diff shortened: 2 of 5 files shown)


wenbingl:
Glad to the effort, does pytest support the Test class as well?

cjermain:
Thanks. Pytest uses functions instead of test classes. [Fixtures in pytest](https://docs.pytest.org/en/latest/fixture.html) allow for setup/teardown of test resources, and can be composed to handle more complex situations. I am most interested in using [parametrized tests](https://docs.pytest.org/en/latest/parametrize.html) so that we can loop over a large number of cases to get better test coverage.

tests/test_layers.py
```diff
@@ -1308,788 +1397,807 @@ def _batch_norm_helper(self, data, gamma, beta, scale, center, axis):
...
-                expected = model.predict(x)
-                self.assertTrue(run_onnx_runtime(onnx_model.graph.name, onnx_model, x, expected, self.model_files))
+        input2 = Input(shape=(None, 5))
+        states = Bidirectional(rnn_class(2, return_state=True))(input1)[1:]
+        out = Bidirectional(rnn_class(2, return_sequences=True))(input2, initial_state=states)
+        model = Model([input1, input2], out)
+        inputs = [x, x]
 
-    def test_rnn_state_passing(self):
-        K.clear_session()
```
cjermain, line 1618: This is being run for all tests now in the `runner` fixture. It helps avoid difficult to reproduce errors that occur due to resource leaking in the tests.

cjermain:
This PR is ready for review. All of the main tests are converted to using pytest. There are a number of additional goals above, but given the diff already, I think they are better left for other PRs. This includes migrating the nightly tests, which still use unittest.

I would highly recommend using the "Hide whitespace changes" to review this PR, since the majority of the changes are indentation.
Screenshot_2020-04-12  WIP  unittest - pytest by cjermain · Pull Request #425 · onnx keras-onnx

@wenbingl & @jiafatom, can you review this PR?

tests/conftest.py
```diff
@@ -0,0 +1,32 @@
+###############################################################################
```
wenbingl, line 1: what's the meaning of 'conftest'?

tests/test_cgan.py
```diff
@@ -120,29 +119,15 @@ def build_discriminator(self):
...
-        onnx_model = keras2onnx.convert_keras(keras_model, keras_model.name)
-        self.assertTrue(run_onnx_runtime(onnx_model.graph.name, onnx_model,
-                                         {keras_model.input_names[0]: x, keras_model.input_names[1]: y}, expected,
-                                         self.model_files))
-
-
-if __name__ == "__main__":
-    unittest.main()
+@pytest.mark.skipif(keras2onnx.proto.tfcompat.is_tf2 and is_tf_keras, reason="Tensorflow 1.x only tests.")
+@pytest.mark.skipif(is_tf_keras and StrictVersion(tf.__version__) < StrictVersion("1.14.0"),
```
wenbingl, line 123: StrictVersion(tensorflow.__version__.split('-')[0]) >= StrictVersi...
otherwise, the whole test cases will failed on tensorflow nightly build.

tests/test_misc.py
```diff
@@ -10,47 +15,41 @@
...
+    converted_model_5 = keras2onnx.convert_keras(model, target_opset=5)
 
-        opset_comparison = converted_model_7.opset_import[0].version > converted_model_5.opset_import[0].version
+    assert converted_model_7 is not None
+    assert converted_model_5 is not None
 
-        self.assertTrue(opset_comparison)
-
-
-if __name__ == '__main__':
```
wenbingl, line 55: we sometimes run this single test file with python cmd line instead of pytest, which helps on debugging. Is there alternative way in the pytest as well?

tests/conftest.py
cjermain: `conftest.py` is the way that `pytest` shares fixtures across the entire test suite.
https://docs.pytest.org/en/latest/fixture.html#conftest-py-sharing-fixture-functions

tests/test_misc.py
cjermain: You can select a specific file by running `pytest` against that file.
```
pytest tests/test_misc.py -vv
```
You can also filter tests for specific strings of interest (e.g. LSTM):
```
pytest tests -k LSTM -vv
```

tests/test_cgan.py
cjermain: I'm confused by what you mean here. The current conditional is the same as that in line 133, so the previous behavior is reproduced. Was the original incorrect?

cjermain:
> Can you run code format to format the code before merge? like 'Reformat code' in pycharm.

I use `vim` and don't have a formatter built in to my editor. There are a number of CLI formatting tools out there, for example [black](https://github.com/psf/black), [yapf](https://github.com/google/yapf), and [autopep8](https://github.com/hhatto/autopep8). Can you be more specific about the formatting you are looking for?

tests/test_cgan.py
wenbingl: The previous code is also ‘imprecise’, but the issue was hidden because this condition wasn't tested due to the first is_tf2 test is already wrong. And it looks the pytest.mark will test both. If you install a tensorflow nightly build. you will see the failure.

wenbingl:
> > Can you run code format to format the code before merge? like 'Reformat code' in pycharm.
> 
> I use `vim` and don't have a formatter built in to my editor. There are a number of CLI formatting tools out there, for example [black](https://github.com/psf/black), [yapf](https://github.com/google/yapf), and [autopep8](https://github.com/hhatto/autopep8). Can you be more specific about the formatting you are looking for?

I saw there were few PEP8 suggestions in my editor tool. But it's not mandatory now. Feel free to ignore it if it is not straightforward in your development.

tests/test_cgan.py
cjermain: Ok, that makes sense. I've made the change.

wenbingl:
Thanks a lot for your contribution!

PR #427, opened 2020-04-06
Fix _create_keras_nodelist for test_rnn_state_passing

jiafatom:
The change in [this](https://github.com/onnx/keras-onnx/pull/417) PR breaks mask_rcnn nightly build, I reverted the relevant part then it breaks `test_rnn_state_passing`, so I find the root cause in `_create_keras_nodelist` and fix it.


wenbingl:
@cjermain , previouslly, I forget where is the code to handle the sharing layer issue. now David has proposed the right way to fix it.

cjermain:
Thanks @jiafatom for resolving that!

PR #439, opened 2020-04-15
Add conversion for FloorDiv, ZerosLike and Fix bug in slice

sonu1-p:
This commit adds model conversion for FloorDiv, ZerosLike

Also fixes a bug in slice conversion. When the begin index is
less than zero, the end index cannot be calculated by adding size.

eg: begin = -1 and size = 1, will give end index = 0 which is incorrect.
The end should be int max in such cases.

Changes to be committed:
	modified:   keras2onnx/_builtin.py


wenbingl:
Thanks for the contribution, it would be nicer to have some unit test, we have some example like test_tf_xxxx in test_layers.py for your reference to add an unit test.

PR #443, opened 2020-04-16
Fix some tf2.x conversion bugs.

wenbingl:


jiafatom:
Did we verify this PR on efficientNet model conversion?

wenbingl:
> Did we verify this PR on efficientNet model conversion?
Yes, it can convert. 
BTW, the EfficientNet doesn't run though the code path which I fixed.

PR #453, opened 2020-04-23
support the random generator ops and fix the issues on tf.op

wenbingl:


wenbingl:
> How can we make sure the correctness of random generator ops? Since it is random, the result can be random for tf ops and onnx ops. Therefore can it guarantee equal?

There is an attr called 'seed'.

PR #459, opened 2020-04-25
Update the behavior on custom op.

wenbingl:


keras2onnx/_tf_ops.py
```diff
@@ -36,11 +36,16 @@ def convert_tf_random_standard_normal(scope, operator, container):
 
 
 def pass_thru_converter(scope, operator, container):
+    """
+    This the converter to copy the original graph node with its def into a ONNX node format.
```
jiafatom, line 40: grammar mistake?

wenbingl: Good catch.

PR #464, opened 2020-04-28
Enable test for efficientNet, fix coverage pytest

jiafatom:


applications/nightly_build/test_efn.py
```diff
@@ -26,6 +25,7 @@ def tearDown(self):
         for fl in self.model_files:
             os.remove(fl)
 
+    @unittest.skip("TODO")
```
wenbingl, line 28: Can you turn this case to tf.keras and validate it as well?

jiafatom: Good point, turn to tf.keras, set rtol=1e-2 atol=5e-4 seems ok.

jiafatom:
> LGTM! How were the tolerance levels in the efficientnet test decided? Are these standard across our model tests?

Decided empirically.

PR #465, opened 2020-04-28
[Test only] other two transformers

jiafatom:


jiafatom:
Slow in nightly build, tested successfully in local dev box.

PR #466, opened 2020-04-28
improve the converter debugging

wenbingl:


keras2onnx/_tf_utils.py
```diff
@@ -39,6 +39,8 @@ def _count_input_nodes(tensor):  # type: (tensorflow.Tensor)->int
         if len(nodes_to_keep) >= _MAX_FOLDING_NODE_NUMBER:
             return -1
 
+        if nd_.type in ['Shape', 'ShapeN', 'Size']:
+            continue
```
jiafatom, line 43: This is the reason that slows down the nightly build (transformers) heavily, please delete these two lines.

wenbingl: Thanks a lot!
The cal_tensor_value is very slow during the conversion, we find a better way to do that.

PR #475, opened 2020-05-04
Handle conv layer spec when input0 is type SpaceToBatchND

jiafatom:


tests/test_layers.py
```diff
@@ -193,15 +193,16 @@ def my_func_2(x):
...
-    model = Sequential()
-    model.add(InputLayer(input_shape=(8, 8, 2)))
-    model.add(keras.layers.DepthwiseConv2D(
-        kernel_size=(3, 3), strides=(1, 1), padding="VALID",
-        data_format='channels_last'))
-    onnx_model = keras2onnx.convert_keras(model, 'test_depthwise_conv2d')
-    data = np.random.rand(3, 8, 8, 2).astype(np.float32)
-    expected = model.predict(data)
-    assert runner('onnx_depthwise_conv2d', onnx_model, data, expected)
+    for use_bias_ in [True, False]:
```
wenbingl, line 196: should it be the test argument?

keras2onnx/_builtin.py
```diff
@@ -118,43 +118,45 @@ def convert_tf_batch_to_space(scope, operator, container):
                                       starts=[0], ends=[1])
         block_prod = oopb.apply_mul(block_size + block_size,
                                     name=operator.full_name + '_mul_0')
-        padded_block_prod = oopb.apply_pad([block_prod,
-                                            ('const_zero_three', oopb.int64, np.array([0, 3], dtype='int64')),
-                                            ('one', oopb.int64, np.array([1], dtype='int64'))],
-                                           name=operator.full_name + '_pad_0')
+        padded_block_prod = oopb.apply_pad(block_prod,
+                                           name=operator.full_name + '_pad_0',
+                                           pads=[0, 3],
```
wenbingl, line 123: where does pads come from?

jiafatom: This is used for division, it is the dimension of the channel, so 0 means N, 3 means C

tests/test_layers.py
jiafatom: done

PR #476, opened 2020-05-07
Fix GPT2 output order mismatch

jiafatom:


applications/nightly_build/test_transformers.py
```diff
@@ -165,6 +165,27 @@ def test_TFBertForQuestionAnswering(self):
...
+            model_list = [TFGPT2Model]
+        # pretrained_weights = 'gpt2'
+        tokenizer_file = 'gpt2_gpt2.pickle'
+        tokenizer = self._get_tokenzier(tokenizer_file)
+        text, inputs, inputs_onnx = self._prepare_inputs(tokenizer)
+        config = GPT2Config()
+        for model_instance_ in model_list:
+            keras.backend.clear_session()
+            model = model_instance_(config)
+            predictions = model.predict(inputs)
```
jiafatom, line 183: This prediction works for transformers 2.5.0, but not for 2.8.0 with the errors:
`>                 batch_element.shape, self.results.shape))
E         ValueError: Mismatch between expected batch size and model output batch size. Output shape = (2, 1, 12, 10, 64), expected output shape = shape (1, 1, 12, 10, 64)
`

PR #478, opened 2020-05-07
Fix GPT2 UT for transformers==2.8.0

jiafatom:
GPT2 UT works for transformers==2.5.0 on my local dev. Nightly build uses transformers 2.8.0 and causes output shape mismatch for keras prediction `model.predict`. This PR fixes it (and it also works for 2.5.0).

The onnx inference result matches `model.predict`, i.e., a list. But need postprocessing for `model(inputs)` because keras outputs one Eagertensor as final result with internal (a tuple of Eagertensors).


applications/nightly_build/test_transformers.py
```diff
@@ -46,11 +48,16 @@ def _get_tokenzier(self, tokenizer_file):
...
         raw_data = json.dumps({
-            'text': 'The quick brown fox jumps over the lazy dog.'
+            'text': 'The quick brown fox jumps over lazy dog.'
         })
         text = json.loads(raw_data)['text']
-        inputs = tokenizer.encode_plus(text, add_special_tokens=True, return_tensors='tf')
-        inputs_onnx = {k_: v_.numpy() for k_, v_ in inputs.items()}
+        # The tokenizers are generated using transformers 2.5.0, but model_max_length is introduced and needed in 2.9.0.
+        if not hasattr(tokenizer, 'model_max_length'):
+            tokenizer.model_max_length = 1024
```
jiafatom, line 56: This change is for transformers 2.9.0 which is just released one hour ago before the change, which breaks nightly build.

PR #480, opened 2020-05-08
Adjust input output sizes when any dim is None

buddhapuneeth:
Whenever inputs and outputs have None as dim (except for batch size), after conversion they will have 0 as the dimension.
Batch size will be converted to 'N'.
input(None, None) -> input(N, 0)

With this fix even other dims will have variables like M1, M2.
input(None, None) -> input(N, M1)

### DIFF
--- a/keras2onnx/_parser_tf.py
+++ b/keras2onnx/_parser_tf.py
@@ -54,6 +54,15 @@ def adjust_input_batch_size(var_type):
     return var_type
 
 
+def adjust_input_output_size(var_type, dim_variable_counter):
+    if len(var_type.shape) > 0:
+        for dim in range(1, len(var_type.shape)):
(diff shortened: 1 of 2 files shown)


buddhapuneeth:
@jiafatom currently if we don't make it M1, M2.. the inputs and outputs where dim = None will be converted to 0. I know the parameterized dimension won't be propagated over the layers. I mean in some cases input's M1 might be same as output's M2. But without this fix, on the **inference side we are having issues as it is expecting dim as 0.**
Simple example (embedded layer) to illustrate this:

  without this fix:
    Embedding_withoutfix
  
   with fix:
Embedding_withfix

Please provide your feedback/suggestion on this.

wenbingl:
Thanks a lot for your contribution!

PR #482, opened 2020-05-08
Support tensorflow 2.2

jiafatom:

PR #484, opened 2020-05-12
Support tensorflow 2.2

jiafatom:
This is from [PR](https://github.com/onnx/keras-onnx/pull/482) with changes addressing reviewer's comments.

### DIFF
--- a/keras2onnx/parser.py
+++ b/keras2onnx/parser.py
@@ -482,20 +482,28 @@ def _get_output_nodes(node_list, node):
     return [n_ for n_ in node_list if n_ not in nodes_has_children]  # need to keep the order.
 
 
+def _filter_out_input(node_name):
+    # tf.keras BN layer sometimes create a placeholder node 'scale' in tf2.x.
+    # Given bn layer will be converted in a whole layer, it's fine to just filter this node out.
+    filter_out = re.match(r"batch_normalization_\d+\/scale$", node_name)
+    filter_out = filter_out or re.match(r"batch_normalization_\d+\/cond", node_name) # since tf 2.2
+    return filter_out
+
+
 def _advance_by_input(cur_node, layer_nodes, subgraph, inputs, graph_inputs, q_overall):
     for input_ in cur_node.inputs:
         predecessor = input_.op
         if is_placeholder_node(predecessor):
-            # tf.keras BN layer sometimes create a placeholder node 'scale' in tf2.x.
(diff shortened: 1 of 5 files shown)


keras2onnx/parser.py
```diff
@@ -482,20 +482,28 @@ def _get_output_nodes(node_list, node):
     return [n_ for n_ in node_list if n_ not in nodes_has_children]  # need to keep the order.
 
 
+def _filter_out_input(node_name):
+    # tf.keras BN layer sometimes create a placeholder node 'scale' in tf2.x. 
+    # Given bn layer will be converted in a whole layer, it's fine to just filter this node out.
+    filter_out = re.match(r"batch_normalization_\d+\/scale$", node_name)
```
wenbingl, line 488: Put all patterns in a list to be easily extended.

keras2onnx/parser.py
```diff
@@ -482,20 +482,28 @@ def _get_output_nodes(node_list, node):
...
+            if not _filter_out_input(predecessor.name):
                 inputs.add(predecessor)
                 graph_inputs.add(predecessor)
+                continue
         if predecessor in layer_nodes or len(layer_nodes) == 0:
             subgraph.append(predecessor)
         else:
-            inputs.add(predecessor)
-            q_overall.put_nowait(predecessor)
+            if not _filter_out_input(predecessor.name):
```
wenbingl, line 504: Can you confirm why some many filters needed?

keras2onnx/parser.py
```diff
@@ -735,7 +743,15 @@ def parse_graph_modeless(topo, graph, target_opset, input_names, output_names, k
 
     for ts_i_ in input_tensors:
         var_type = _adjust_input_batch_size(infer_variable_type(ts_i_, target_opset))
-        str_value = ts_i_.name
+        if ts_i_.name.endswith(':0'):
```
wenbingl, line 746: Any special things in the model causes ':0'?

keras2onnx/parser.py
jiafatom: This seems redundant, removed.

keras2onnx/parser.py
jiafatom: This is tensor name so it has ":0". This code path is `parse_graph_modeless`, in the traditional path, we also have logic to handle this, see [here](https://github.com/onnx/keras-onnx/blob/89b0c459ea2dd1aa086d16983d70547992ef8095/keras2onnx/parser.py#L775-L785)

Issue #505, opened 2020-05-31
converting Relu6 


ShiraLightricks:
Hello,
I'm trying to convert my keras model to onnx, using keras2oonx.
python=3.7, keras2onnx=1.7, tensorflow=2.2.0, onnx=1.5

I get this warning:
WARN: No corresponding ONNX op matches the tf.op node model/up_5_block_3X3.2/Relu6 of type Relu6
The generated ONNX model needs run with the custom op supports.

I saw that this issue with the relu6 was resolved in the keras2onnx version 1.6.5 in the file: keras2onnx.ke2onnx.activations,
but when I run the code I see it doesn't use this function, and I didn't find an explanation on how to use it, or how to implement my own custom convert.
Thanks!
Shira

jiafatom:
Just checked in the [fix](https://github.com/onnx/keras-onnx/pull/506), please pull latest keras2onnx master and onnxconverter-common master and try again.

ShiraLightricks:
Hi,
Thanks for the quick answer, it worked but after I changed something in keras2onnx._builtin.py:
in the function "convert_tf_relu6" raw_operator didn't have an attribute "input", it had "inputs" (or "inputs_dtype" which is a list) and I changed to:
dtype=operator.raw_operator.inputs[0].dtype.as_numpy_dtype

Also I just want to ask about the keras2onnx.ke2onnx.activations.py file - I don't see where in the code it is used or how.
Thank you!
Shira

jiafatom:
Could you provide a simple example to cover your usage, so we can include it in our unit test?

For `keras2onnx.ke2onnx.activations.py`, it has `activation_map` which maps the activation op to the conversion function. It also has the conversion function `convert_keras_activation` to convert keras activation layer, which will be globally registered to convert keras layers. Search for these in our code base, and you will find usage.

Previously we only handle keras layer relu6 activation but not tf.nn.relu6, now we added it per your issue.

ShiraLightricks:
I'm not sure how to give a simple example, 
I have a very complex network and I use the Relu6 in several layers, but thanks for the fix!

PR #510, opened 2020-06-01
Handle different input cases for tf.nn.relu6 support

jiafatom:
The [issue](https://github.com/onnx/keras-onnx/issues/505) mentions it needs additional handling for input, which is different from our UT. Add the case here, will ask the issue owner for UT.

### DIFF
--- a/keras2onnx/_builtin.py
+++ b/keras2onnx/_builtin.py
@@ -10,6 +10,7 @@ import numpy as np
 
 from keras2onnx._consts import TYPES, NCHW_TO_NHWC, NHWC_TO_NCHW, HWCN_TO_NCHW
 from onnx import numpy_helper
+from onnx.mapping import TENSOR_TYPE_TO_NP_TYPE
 from .common.utils import count_dynamic_dim
 from .common.onnx_ops import apply_identity, apply_reshape, OnnxOperatorBuilder
 from .funcbook import converter_func, set_converters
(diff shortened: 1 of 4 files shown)


keras2onnx/_builtin.py
```diff
@@ -1726,11 +1726,13 @@ def convert_tf_read_variable_op(scope, operator, container):
...
 def convert_tf_relu6(scope, operator, container):
     oopb = OnnxOperatorBuilder(container, scope)
+    operator_input = operator.raw_operator.input if hasattr(operator.raw_operator, 'input') else \
+        operator.raw_operator.inputs[0]
     oopb.apply_op_with_output("apply_relu6",
                               operator.input_full_names,
                               operator.output_full_names,
                               name=operator.full_name + '_clip',
-                              dtype=operator.raw_operator.input.dtype.as_numpy_dtype)
+                              dtype=operator_input.dtype.as_numpy_dtype)
```
wenbingl, line 1735: dtype supposes to be stored in a variable of scope, why is it different here?

jiafatom: The issue owner mentions there is some case that `operator.raw_operator` does not have attribute 'input', only has 'inputs[0]' instead.

wenbingl: I see, the previous apply_clip implementation is so bad. can you avoid to call this function?

jiafatom: synced offline, get the input type from `operator`, not from `raw_operator`

Issue #511, opened 2020-06-02
convert softmax


ShiraLightricks:
Hi,
I have a problem with the softmax activation layer, it seems the code is supposed to handle this case but it gives me the following error:

Unsupported activation in the layer(<function softmax at 0x1365b6320>)

details:
python=3.7, keras2onnx=1.7, tensorflow=2.2.0, onnx=1.5

Thanks!
Shira

jiafatom:
Please give more information about the softmax layer, is it a tf layer? Since we don't have the model, can you give the stack trace?

shonigs:
It seems that this doesn't work  
`model.layers[-1].activation = tf.keras.activations.softmax`
and this does 
```
model.layers[-1].activation = tf.keras.activations.linear
output = tf.keras.layers.Softmax()(model.output)
model=tf.keras.Model(inputs=model.inputs, outputs=output)`
```

jiafatom:
Fixed [here](https://github.com/onnx/keras-onnx/pull/513), please pull keras2onnx master and try again.

PR #513, opened 2020-06-03
Add both tf.nn.X and tf.compat.v1.nn.X to activation_map

jiafatom:
The [issue](https://github.com/onnx/keras-onnx/issues/511) cannot convert softmax activation because `operator.raw_operator.activation` is a function that is not the same as `tf.nn.softmax` or `keras.activations.get('softmax')`. We need import `tf` from `keras2onnx.proto.tfcompat`

### DIFF
--- a/keras2onnx/ke2onnx/activation.py
+++ b/keras2onnx/ke2onnx/activation.py
@@ -34,7 +34,16 @@ activation_map = {activation_get('sigmoid'): apply_sigmoid,
                   tf.nn.relu: apply_relu,
                   tf.nn.relu6: apply_relu_6,
                   tf.nn.elu: apply_elu,
-                  tf.nn.tanh: apply_tanh}
+                  tf.nn.selu: apply_selu,
+                  tf.nn.tanh: apply_tanh,
+                  'sigmoid': apply_sigmoid,
+                  'softmax': apply_softmax,
+                  'linear': apply_identity,
+                  'relu': apply_relu,
+                  'elu': apply_elu,
+                  'selu': apply_selu,
+                  'tanh': apply_tanh,
+                  'hard_sigmoid': apply_hard_sigmoid}
 
 
 def convert_keras_activation(scope, operator, container):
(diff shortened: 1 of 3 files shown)


keras2onnx/ke2onnx/activation.py
```diff
@@ -34,7 +34,16 @@
                   tf.nn.relu: apply_relu,
                   tf.nn.relu6: apply_relu_6,
                   tf.nn.elu: apply_elu,
-                  tf.nn.tanh: apply_tanh}
+                  tf.nn.selu: apply_selu,
```
wenbingl, line 37: any idea why the function is not as tf.nn? if it was customized function was created by the user, it will be pretty dangerous to  use the name.

jiafatom: Through debugging, I can only see the function has different address from tf.nn.softmax.

jiafatom: re-checked, code updated.

keras2onnx/ke2onnx/activation.py
```diff
@@ -4,7 +4,7 @@
 # license information.
 ###############################################################################
 import numpy as np
-import tensorflow as tf
+from keras2onnx.proto.tfcompat import tensorflow as tf
```
wenbingl, line 7: good finding, you could be add both tensorflow.compat.v1.nn and tensorflow.nn

jiafatom: good, updated.

keras2onnx/ke2onnx/activation.py
```diff
@@ -34,7 +34,15 @@
                   tf.nn.relu: apply_relu,
                   tf.nn.relu6: apply_relu_6,
                   tf.nn.elu: apply_elu,
-                  tf.nn.tanh: apply_tanh}
+                  tf.nn.selu: apply_selu,
+                  tf.nn.tanh: apply_tanh,
+                  tf.compat.v1.nn.sigmoid: apply_sigmoid,
```
wenbingl, line 39: need check tf version before add it.

jiafatom: checking hasattr(tf.compact, 'v1')

PR #514, opened 2020-06-04
Support tf.nn.leaky_relu and fix advanced_activations

jiafatom:
Need check in onnxconverter-common [change](https://github.com/microsoft/onnxconverter-common/pull/117) first.

### DIFF
--- a/keras2onnx/ke2onnx/activation.py
+++ b/keras2onnx/ke2onnx/activation.py
@@ -6,8 +6,8 @@
 import numpy as np
 import tensorflow as tf
 from ..proto import keras, is_tf_keras
-from ..common.onnx_ops import apply_elu, apply_hard_sigmoid, apply_relu, apply_relu_6, apply_sigmoid, apply_tanh, \
-    apply_softmax, apply_identity, apply_selu, apply_mul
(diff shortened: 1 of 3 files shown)


keras2onnx/ke2onnx/activation.py
```diff
@@ -51,29 +53,37 @@ def convert_keras_activation(scope, operator, container):
...
+    elif activation_type == keras.layers.advanced_activations.PReLU:
+        apply_prelu(scope, input_name, output_name, container, slope=operator.raw_operator.get_weights()[0])
+    elif activation in [relu6] or (hasattr(activation, '__name__') and activation.__name__ == 'relu6'):
         # relu6(x) = min(relu(x), 6)
         np_type = TENSOR_TYPE_TO_NP_TYPE[operator.inputs[0].type.to_onnx_type().tensor_type.elem_type]
         zero_value = np.zeros(shape=(1,), dtype=np_type)
         apply_relu_6(scope, input_name, output_name, container,
                      zero_value=zero_value)
-    elif activation.__name__ in ['swish']:
+    elif hasattr(activation, '__name__') and activation.__name__ == 'swish':
```
wenbingl, line 86: I think this change will break swish, since you get type of activation, i guess the __name__ will become 'function' instead of 'swish'.

jiafatom: this is still `activation` not `activation_type` for `swish`.

PR #527, opened 2020-06-18
Pin keras==2.3.1 for tensorflow 1.x in nightly build

jiafatom:
keras 2.4.0 released to pypi today (6/17/2020) which installs tensorflow 2.x, then it breaks nightly build for tf 1.x case. Pin keras==2.3.1 accordingly.


wenbingl:
Should the tensorflow be installed after keras to override its dependency.

jiafatom:
> Should the tensorflow be installed after keras to override its dependency.

Just tried but not work, having the error: ERROR: keras 2.4.0 has requirement tensorflow>=2.2.0, but you'll have tensorflow 2.0.0 which is incompatible.

PR #532, opened 2020-06-22
Add support for ReverseV2 operator

buddhapuneeth:
Reverse V2 is not an ONNX operator, so it is translated into Reverse Sequence and other set of valid ONNX operators.

### DIFF
--- a/keras2onnx/_builtin.py
+++ b/keras2onnx/_builtin.py
@@ -2132,6 +2132,60 @@ def convert_tf_zeros_like(scope, operator, container):
                               name=operator.full_name)
 
 
+@converter_func(TYPES.ReverseV2)
+def convert_tf_reverse(scope, operator, container):
+    node = operator.raw_operator
+    oopb = OnnxOperatorBuilder(container, scope)
+    ip = node.inputs[0]
+    axes = _cal_tensor_value(node.inputs[1])
+
+    len_axes = len(axes)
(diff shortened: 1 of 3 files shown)


buddhapuneeth:
Can one of you check why few checks are failing? Batch_norm test is failing, which is not related to my changes and unable to reproduce locally.

jiafatom:
> Can one of you check why few checks are failing? Batch_norm test is failing, which is not related to my changes and unable to reproduce locally.

Just checked there is one test failure -- random number fails -- Sometimes it happens. Could you make a trivial change to your PR and resubmit again? It should be fine.

keras2onnx/_builtin.py
```diff
@@ -2211,6 +2211,61 @@ def convert_tf_zeros_like(scope, operator, container):
                               name=operator.full_name)
 
 
+@converter_func(TYPES.ReverseV2)
+def convert_tf_reverse(scope, operator, container):
+    node = operator.raw_operator
+    oopb = OnnxOperatorBuilder(container, scope)
+    ip = node.inputs[0]
+    axes = _cal_tensor_value(node.inputs[1])
+    if len(_cal_tensor_shape(axes)) != 1:
```
jiafatom, line 2220: Combine these two if?

buddhapuneeth: done

PR #536, opened 2020-06-24
Convert keras Dense via Gemm if possible

jiafatom:
Previously keras Dense is converted to MatMul + Add, with activation. It is more efficient to fuse MatMul + Add into Gemm if some criteria meet.


wenbingl:
> Previously keras Dense is converted to MatMul + Add, with activation. It is more efficient to use Gemm if some criteria meet.

Should it be done by the optimizer???

PR #545, opened 2020-06-30
Add SelectV2 conversion

affranchi:
This is the change to convert "SelectV2" into "Where" operator.

### DIFF
--- a/keras2onnx/_builtin.py
+++ b/keras2onnx/_builtin.py
@@ -1367,6 +1367,36 @@ def convert_tf_select(scope, operator, container):
                                   op_version=9)
 
 
+@converter_func(TYPES.SelectV2)
+def convert_tf_selectv2(scope, operator, container):
+    if operator.target_opset < 9:
+        raise ValueError("SelectV2 op is not supported for opset = " + str(operator.target_opset))
+    else:
+        oopb = OnnxOperatorBuilder(container, scope)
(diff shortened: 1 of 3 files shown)


keras2onnx/_builtin.py
```diff
@@ -1473,6 +1473,36 @@ def convert_tf_select(scope, operator, container):
                                   op_version=9)
 
 
+@converter_func(TYPES.SelectV2)
+def convert_tf_selectv2(scope, operator, container):
+    if operator.target_opset < 9:
```
wenbingl, line 1478: Since oopb.add_node_with_output already specifies the opset, it is not needed to add an opset version check here and the converter framework will take care of it.

wenbingl:
check here: https://github.com/onnx/keras-onnx/blob/b3d6b6486fe56640c48c62dd098e9405e35b4e9f/keras2onnx/_tf_ops.py#L41
The selectV2 already was implemented, What's the error you encountered?

affranchi:
> check here:
> 
> https://github.com/onnx/keras-onnx/blob/b3d6b6486fe56640c48c62dd098e9405e35b4e9f/keras2onnx/_tf_ops.py#L41
> 
> 
> The selectV2 already was implemented, What's the error you encountered?

Thank you for pointing out. My version didn't have it when I worked on.
Closed it.

Thanks,
Kenny

PR #548, opened 2020-07-02
Use oopb.apply_reshape in convert_keras_reshape in main.py

jiafatom:
Previously `convert_keras_reshape` directly `apply_reshape` using name `operator.raw_operator.name`, which does not guarantee the unique operator name in the scope. It results in two `Reshape` nodes with identical name. Use `oopb` which solves this issue internally.
Now effdet can be successfully converted and tested with some few changes (activation) in the original model.

Update: the reason why it has two identical names is because the bug in `interim.py`, as I explained below. Fix that is enough. Well, it is also fine to use `oopb` in `main.py`.

### DIFF
--- a/applications/nightly_build/test_keras_applications.py
+++ b/applications/nightly_build/test_keras_applications.py
@@ -80,6 +80,23 @@ class TestKerasApplications(unittest.TestCase):
         res = run_image(model, self.model_files, img_path)
         self.assertTrue(*res)
 
+    def test_effdet(self):
+        from model import efficientdet
+        m_train, model = efficientdet(phi=0,
+                                      weighted_bifpn=True,
+                                      num_classes=90,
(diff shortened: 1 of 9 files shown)


jiafatom:
> direct apply_func should work as well,
> please check here: https://github.com/onnx/keras-
> onnx/blob/5964f90c22b80a0df8d35524df477855c6cd5aa2/keras2onnx/common/interim.py#L125, can you check what happened?

This piece of code has bug:
Eff-det has a node name `reshape`, and it goes through `cur_count` 5 times, so we have onnx node `reshape`, `reshape_1`, ..., `reshape_4`. And eff-det has another node `reshape_1`, it will not go to `if attrs['name'] in self.node_names:`, so it will have another `reshape_1`. We should change the code here.

PR #549, opened 2020-07-02
Add the user specified input type in the converter api

wenbingl:


jiafatom:
Can we have a unit test for this change?

wenbingl:
> Can we have a unit test for this change?

Yes, working on it.

PR #554, opened 2020-07-07
Allow softsign in recurrent layers

jwfromm:
Extremely minimal PR that allows softsign activations in recurrent layers. Please let me know what tests, if any, need to be added.


jiafatom:
Thanks for your contribution!

PR #567, opened 2020-07-21
parser: filter the input "conv_7b_bn/cond/input"

yorkie:
This is to fix an error when converting an inception_resnet_v2 Keras model to ONNX:

```
AssertionError: conv_7b_bn/cond/input_0:01 is disconnected, check the parsing log for more details.
```

After investigating some background, I found inception_resnet_v2 has a name "conv_7b_bn" for a BN layer, just fixes by filtering as we do filter for "batch_normalization".

See [tensorflow/python/keras/applications/inception_resnet_v2.py](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/applications/inception_resnet_v2.py#L203), which defined a custom `conv_7b_bn` for a specific BN layer.

### DIFF
--- a/applications/nightly_build/test_keras_applications_v2.py
+++ b/applications/nightly_build/test_keras_applications_v2.py
@@ -58,7 +58,7 @@ class TestKerasApplications(unittest.TestCase):
         self.assertTrue(*res)
     
     def test_InceptionResNetV2(self):
-        InceptionResNetV2 = keras.applications.inception_resnet_v2
+        InceptionResNetV2 = keras.applications.inception_resnet_v2.InceptionResNetV2
         model = InceptionResNetV2(include_top=True)
         res = run_image(model, self.model_files, img_path, target_size=299, tf_v2=True)
         self.assertTrue(*res)
--- a/applications/nightly_build/test_keras_applications_v2.py
+++ b/applications/nightly_build/test_keras_applications_v2.py
@@ -56,6 +56,12 @@ class TestKerasApplications(unittest.TestCase):
         model = InceptionV3(include_top=True)
         res = run_image(model, self.model_files, img_path, target_size=299, tf_v2=True)
         self.assertTrue(*res)
+    
+    def test_InceptionResNetV2(self):
+        InceptionResNetV2 = keras.applications.inception_resnet_v2
+        model = InceptionResNetV2(include_top=True)
+        res = run_image(model, self.model_files, img_path, target_size=299, tf_v2=True)
+        self.assertTrue(*res)
 
     def test_ResNet50(self):
         ResNet50 = keras.applications.resnet_v2.ResNet50V2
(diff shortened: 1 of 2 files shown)


jiafatom:
Thanks for your contribution! Can you add this model conversion into our unit test? Is it for tf2, then we can put it into `test_keras_applications_v2.py`?
About `conv_7b_bn/cond/input`, would `7` just be index number, then we should use regular expression to express that, right?

yorkie:
> Thanks for your contribution! Can you add this model conversion into our unit test? Is it for tf2, then we can put it into `test_keras_applications_v2.py`?
> About `conv_7b_bn/cond/input`, would `7` just be index number, then we should use regular expression to express that, right?

@jiafatom Thanks for your sweet guide, as for the 7n, it's hardcoded at https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/applications/inception_resnet_v2.py#L203, I'm not sure if it's an index, but I could try to match with number :)

jiafatom:
> > Thanks for your contribution! Can you add this model conversion into our unit test? Is it for tf2, then we can put it into `test_keras_applications_v2.py`?
> > About `conv_7b_bn/cond/input`, would `7` just be index number, then we should use regular expression to express that, right?
> 
> @jiafatom Thanks for your sweet guide, as for the 7n, it's hardcoded at https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/applications/inception_resnet_v2.py#L203, I'm not sure if it's an index, but I could try to match with number :)

We can do a quick experiment to verify this. Just remove the hardcoded name `conv_7b`, then it will generate a name for it. Then we can see how the name looks like.

yorkie:
@jiafatom Do I need to add a new test `test_keras_applications_v2.py`? I don't find it at https://github.com/onnx/keras-onnx/tree/master/tests :(

jiafatom:
> @jiafatom Do I need to add a new test `test_keras_applications_v2.py`? I don't find it at https://github.com/onnx/keras-onnx/tree/master/tests :(

Just add the test [here](https://github.com/onnx/keras-onnx/blob/master/applications/nightly_build/test_keras_applications_v2.py) in our nightly build, thanks.

yorkie:
@jiafatom I have fixed the nits that you picked up :)

jiafatom:
> @jiafatom I have fixed the nits that you picked up :)

When I run the nightly build, it throws the error:
```python
>       model = InceptionResNetV2(include_top=True)
E       TypeError: 'module' object is not callable
```
See the build error [here](https://dev.azure.com/onnxmltools/ketone/_build/results?buildId=14228&view=logs&j=7b0f2069-5587-5cdf-c0b7-b313d0876a05&t=93f479da-a728-5bc1-3564-eeec6f964683)

applications/nightly_build/test_keras_applications_v2.py
```diff
@@ -56,6 +56,12 @@ def test_InceptionV3(self):
         model = InceptionV3(include_top=True)
         res = run_image(model, self.model_files, img_path, target_size=299, tf_v2=True)
         self.assertTrue(*res)
+    
+    def test_InceptionResNetV2(self):
+        InceptionResNetV2 = keras.applications.inception_resnet_v2
```
jiafatom, line 61: InceptionResNetV2 = keras.applications.InceptionResNetV2?

yorkie: It should be `keras.applications.inception_resnet_v2. InceptionResNetV2`, sorry about that.

PR #574, opened 2020-07-27
Convert tf.pow to Identitiy for pow=1

jiafatom:


wenbingl:
It looks pretty trivial, why are you doing this?

PR #581, opened 2020-08-01
Support tf MirrorPad conversion

jiafatom:
Converting a deblur GAN model [here](https://github.com/RaphaelMeudec/deblur-gan) and see a `MirrorPad` op. Seems the current ORT only supports Pad reflect mode until 3 dim. For 4 dim (a, b, c, d), the result mismatch for d > 1 case.

### DIFF
--- a/keras2onnx/_tf_utils.py
+++ b/keras2onnx/_tf_utils.py
@@ -81,7 +81,7 @@ def tf_attrs_to_onnx(node):
         if s_.startswith('T'):  # all T starts attr is TF internal.
             continue
         v = node.get_attr(s_)
-        if isinstance(v, tensorflow.dtypes.DType):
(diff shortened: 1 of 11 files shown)


keras2onnx/_tf_utils.py
```diff
@@ -81,7 +81,7 @@ def tf_attrs_to_onnx(node):
         if s_.startswith('T'):  # all T starts attr is TF internal.
             continue
         v = node.get_attr(s_)
-        if isinstance(v, tensorflow.dtypes.DType):
+        if hasattr(tensorflow.dtypes, 'DType') and isinstance(v, tensorflow.dtypes.DType):
```
jiafatom, line 84: For python 3.5 CI build, we see `AttributeError: module 'tensorflow.dtypes' has no attribute 'DType'` for this newly added UT.

PR #589, opened 2020-08-07
Add models into nightly build (week of 20200804)

jiafatom:


applications/nightly_build/test_ssrnet.py
```diff
@@ -0,0 +1,232 @@
...
+class TestSSRNet(unittest.TestCase):
+
+    def setUp(self):
+        self.model_files = []
+
+    def tearDown(self):
+        for fl in self.model_files:
+            os.remove(fl)
+
+    @unittest.skipIf(not is_weekend,
```
wenbingl, line 215: what do you think the weekend is different from weekday, regarding to running the tests?

jiafatom: On weekdays, we sometimes need test nightly build for import PR changes, we don't want to take too long but cover important models. On weekend, we can test even more (like these models) and spend more time.

wenbingl: try the pipeline variable for the manual check:
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch

PR #610, opened 2020-09-07
Support for recurrent_v2 layers for TF 2.0

### LINKED ISSUE #607: TF v2 support for Recurrent/Bidirectional
@DronistB posted an issue on #413 showing that the new TensorFlow V2 API for Keras does not work with the existing conversion code for Bidirectional RNN layers. This occurs because TensorFlow's new RNN layers are in a different module (`tensorflow.python.keras.layers.recurrent_v2`). At a minimum, the following code will need to change to support the TensorFlow versions as well as the conventional Keras classes.

https://github.com/onnx/keras-onnx/blob/ff17787c393e2ce34d43185447d7354525f3ba87/keras2onnx/ke2onnx/bidirectional.py#L25-L37

Happy to take a look at resolving this when I have bandwidth.

cjermain:
In #607, it was observed that TensorFlow versions before 2.0 can have problems converting TF 2.0 compatible models because the classes are expected to be the previous API. For example in TF 1.14 the v2 compatible classes are present in `tensorflow.keras.layers.recurrent_v2`, but do not match `tensorflow.keras.layers`. This PR resolves #607 by adding support for these layers. For versions of TensorFlow that are above 2.0, this is not a problem because `tensorflow.keras.layers` uses the v2 layers.

This strategy is tested by expanding the `RNN_CLASSES` list in the tests, so that `recurrent_v2` classes are tested if present.

### DIFF
--- a/keras2onnx/ke2onnx/bidirectional.py
+++ b/keras2onnx/ke2onnx/bidirectional.py
@@ -6,8 +6,11 @@
 import collections
 import numbers
 from ..common import cvtfunc
-from ..proto import keras
+from ..proto import keras, is_tf_keras, is_tensorflow_later_than
 from . import simplernn, gru, lstm
+
+LSTM_CLASSES = {keras.layers.LSTM}
+GRU_CLASSES = {keras.layers.GRU}
 
 
 def _calculate_keras_bidirectional_output_shapes(operator):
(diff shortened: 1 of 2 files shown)


cjermain:
@wenbingl, it looks like the linux-conda-ci build is still waiting to start. Any ideas on how to resolve it?

wenbingl:
> @wenbingl, it looks like the linux-conda-ci build is still waiting to start. Any ideas on how to resolve it?

this is the recent random failure of azp, we don't have a good way to fix it.

And recently, ONNX sig add a DCO requirement which you needs sign your commit. After you sign the commit, I can try to manually run the checker and forcedly merge it if the azp issue still exists.

cjermain:
Ok, got it. I re-committed and force-pushed the change since it was buried under a few merges.

wenbingl (approved):

jiafatom:
This PR fails on linux-tf-keras CI build for tensorflow==1.15.0. The fix is [here](https://github.com/onnx/keras-onnx/pull/661).

PR #614, opened 2020-09-21
Support TimeDistributed dynamic input shape and more Bidirectional merge_modes

jiafatom:
Through this change, we converted 3 more keras models:
one from [repo](https://github.com/kamalkraj/Named-Entity-Recognition-with-Bidirectional-LSTM-CNNs/), the other two from [repo](https://github.com/rolczynski/Automatic-Speech-Recognition)

### DIFF
--- a/keras2onnx/parser.py
+++ b/keras2onnx/parser.py
@@ -116,8 +116,15 @@ def _on_parsing_time_distributed_layer(graph, node_list, layer, model, varset, p
     k2o_logger().debug('td_layer output: ' + oname)
     o1 = varset.get_local_variable_or_declare_one(oname, infer_variable_type(o_, varset.target_opset))
     oshapes1 = [-1 if s_ is None else s_ for s_ in oshapes[0]]
-    operator_reshape_1 = varset.declare_local_operator(TYPES.TD_Reshape,
-                                                       op_name=layer.name + '_reshape_1', target_shape=oshapes1)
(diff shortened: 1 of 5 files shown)


keras2onnx/parser.py
```diff
@@ -116,8 +116,15 @@ def _on_parsing_time_distributed_layer(graph, node_list, layer, model, varset, p
     k2o_logger().debug('td_layer output: ' + oname)
     o1 = varset.get_local_variable_or_declare_one(oname, infer_variable_type(o_, varset.target_opset))
     oshapes1 = [-1 if s_ is None else s_ for s_ in oshapes[0]]
-    operator_reshape_1 = varset.declare_local_operator(TYPES.TD_Reshape,
-                                                       op_name=layer.name + '_reshape_1', target_shape=oshapes1)
+    num_minus_1 = len(list(i for i in oshapes1 if i is not None and i < 0))
+    if num_minus_1 > 1:
+        operator_reshape_1 = varset.declare_local_operator(TYPES.TD_Reshape,
```
wenbingl, line 121: [2:]?
the first 2 dims can be removed?

jiafatom: No, it is not removed. In this case `input_name` contains info for first 2 dim, and it will slice and concat with [2:] inside `convert_reshape_timedistributed`

wenbingl (approved):

PR #668, opened 2020-12-17
Set axes=[0] for apply_squeeze rather than leaving unset

jiafatom:
We should set axes=[0], because Squeeze op has different meaning if we don't set axes -- it squeezes out all dimensions that has shape dim 1.

### DIFF
--- a/keras2onnx/_builtin.py
+++ b/keras2onnx/_builtin.py
@@ -553,6 +553,10 @@ def convert_tf_depth_to_space(scope, operator, container):
     node = operator.raw_operator
     block_size = node.get_attr('block_size')
     oopb = OnnxOperatorBuilder(container, scope)
+    if container.target_opset < 13:
+        op_version = 11
+    else:
+        op_version = 13
     if _is_nhwc(node):
         adjusted_input_name = oopb.apply_transpose(operator.input_full_names,
                                                    name=operator.full_name + '_pre_transpose',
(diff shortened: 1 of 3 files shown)


keras2onnx/_builtin.py
```diff
@@ -553,6 +553,10 @@ def convert_tf_depth_to_space(scope, operator, container):
     node = operator.raw_operator
     block_size = node.get_attr('block_size')
     oopb = OnnxOperatorBuilder(container, scope)
+    if container.target_opset < 13:
```
wenbingl, line 556: which op will use this op_version? should we check the target_opset lower bound as well?

keras2onnx/_builtin.py
```diff
@@ -2360,10 +2380,14 @@ def convert_tf_variable_v2(scope, operator, container):
 def convert_tf_where(scope, operator, container):
     oopb = OnnxOperatorBuilder(container, scope)
     node = operator.raw_operator
+    if container.target_opset < 13:
+        op_version = 9
+    else:
+        op_version = 13
```
wenbingl, line 2386: it looks there is no difference on using opset 11 or opset 13. Conventionally,  the converted model will use the minimal opset number, instead of the opset specified by the user.

keras2onnx/_builtin.py
jiafatom: synced offline.

keras2onnx/_builtin.py
jiafatom: synced offline.

wenbingl (approved):

PR #689, opened 2021-02-01
Pin numpy<=1.19.5 in requirement_dev

jiafatom:
Fix nightly build failure since 1/30/2021. For numpy 1.20.0:
DeprecationWarning: `np.object` is a deprecated alias for the builtin `object`. To silence this warning, use `object` by itself. Doing this will not modify any behavior and is safe. 
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
Then it throws errors when defining keras models.
So use numpy 1.19.5 in our nightly build.


jiafatom:
I manually start the UT, and all passed.

wenbingl (approved):

PR #705, opened 2021-04-09
Add test two zeropadding

jiafatom:


wenbingl (approved):

jiafatom:
The UT example is referred here: https://github.com/microsoft/onnxconverter-common/issues/176

